PR #162, opened 2015-08-31
add listener for iFrame message so callback is executed when the page returns with the token

jnfsmile:


weijjia:
Can you tell me a little bit more on why you want to do this?

jnfsmile:
The iframe functionality isn't working with the current version...

weijjia:
What do you mean by it's not working with the current version? Do you have any logs to show me what's actually going wrong?

jnfsmile:
When calling acquireToken for the accessToken, no result is returning. The system works only for idToken.
My fix enables returning an accessToken after idToken is obtained.

weijjia:
@jnfsmile Well, I don't believe so. Could you send me a fiddler trace show that it doesn't work when calling acquireToken for the access token?

tushargupta51:
@jnfsmile I am not sure I understand the error scenario you explained. AcquireToken method renews an id_token for app's backend (if the Web API, client wants to access lives in the same domain as client), otherwise it renews an access_token for an external web api. The token is renewed using an iframe, AAD returns back the token to `redirect_uri` specified in the original request which is app's root by default. The token is then extracted and saved in the cache.

jnfsmile:
The returned value isn't caught and used

tushargupta51:
@jnfsmile If you want to use the returned token, it is available in the callback method that you pass to acquireToken or RenewToken method. By default, library caches the token in local/session storage.

tushargupta51:
Closing this PR as the registered callback is called with the returned token in the handleWindowCallback method. Callback is specified at the time of calling acquireToken method. Please let us know if you have questions.

PR #205, opened 2015-12-17
Redirct Params UI Router

jasteph:
fix: Navigating to private state with route parameters

Every time a user changes states, if they are required to login, we
record the url parameters and save them locally. On return we check to
see if there are any route params and redirect the user appropriately
using them.

### DIFF
--- a/lib/adal-angular.js
+++ b/lib/adal-angular.js
@@ -120,13 +120,16 @@ if (typeof module !== 'undefined' && module.exports) {
                                         // redirect to login requested page
                                         var loginStartPage = _adal._getItem(_adal.CONSTANTS.STORAGE.START_PAGE);
                                         if (loginStartPage) {
+                                            // Check to see if any params were stored
                                             var paramsJSON = _adal._getItem(_adal.CONSTANTS.STORAGE.START_PAGE_PARAMS); 
                                             
                                             if (paramsJSON) {
+                                                // If params were stored redirect to the page and then 
+                                                // initialize the params
                                                 var loginStartPageParams = JSON.parse(paramsJSON);
-                                                $location.path(loginStartPage).search(loginStartPageParams);
+                                                $location.url(loginStartPage).search(loginStartPageParams);
                                             } else {
-                                                $location.path(loginStartPage);
+                                                $location.url(loginStartPage);
                                             }
                                         }
                                     }, 1);
@@ -163,12 +166,12 @@ if (typeof module !== 'undefined' && module.exports) {
                 };
 
                 var loginHandler = function () {
-                    _adal.info('Login event for:' + $location.$$path);
+                    _adal.info('Login event for:' + $location.$$url);
                     if (_adal.config && _adal.config.localLoginUrl) {
                         $location.path(_adal.config.localLoginUrl);
                     } else {
                         // directly start login flow
-                        _adal._saveItem(_adal.CONSTANTS.STORAGE.START_PAGE, $location.$$path);
+                        _adal._saveItem(_adal.CONSTANTS.STORAGE.START_PAGE, $location.$$url);
                         _adal.info('Start login at:' + window.location.href);
                         $rootScope.$broadcast('adal:loginRedirect');
                         _adal.login();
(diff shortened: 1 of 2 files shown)


lib/adal-angular.js
```diff
@@ -175,16 +185,25 @@ if (typeof module !== 'undefined' && module.exports) {
...
-                            _adal.info('Route change event for:' + $location.$$path);
+                            _adal.info('Route change event for:' + $location.$$url);
                             loginHandler();
                         }
                     }
                 };
 
-                var stateChangeHandler = function (e, nextRoute) {
-                    if (nextRoute && isADLoginRequired(nextRoute, _adal.config)) {
+                var stateChangeHandler = function (e, toState, toParams, fromState, fromParams){
```
weijjia: can you also share the ui-route config? my main concern is this will change how people use us today, and this is definitely not what we want.

weijjia:
Thanks for the PR. This only has the change on angular level, if we support it, we should also make the change on the plain js level. Can you also make this change?

lib/adal-angular.js
laughin1:
```
/** @ngInject */
export function routerConfig($stateProvider: ng.ui.IStateProvider, $urlRouterProvider: ng.ui.IUrlRouterProvider) {
    $stateProvider
		.state('home', {
			url: '/',
            views: {
                '': {
                    templateUrl: 'app/home/home.view.html',
                    controller: 'HomeController',
                    controllerAs: 'home'
                }
			}
        })
        .state('entities', {
            abstract: true, // Set this to true so it will never be activated directly
            url: '',
            templateUrl: 'app/main/coraMain.view.html',
            controller: 'CoraMainController',
            controllerAs: 'main'
        })
        .state('secret', <any>{
            url: '/secret?param1&param2&param3',
            templateUrl: 'app/secret.view.html',
            controller: 'SecretController',
            controllerAs: 'secret',
            requireADLogin: true
        });

	$urlRouterProvider.otherwise('/');
}
```

jasteph:
```
/** @ngInject */
export function routerConfig($stateProvider: ng.ui.IStateProvider, $urlRouterProvider: ng.ui.IUrlRouterProvider) {
    $stateProvider
		.state('home', {
			url: '/',
            views: {
                '': {
                    templateUrl: 'app/home/home.view.html',
                    controller: 'HomeController',
                    controllerAs: 'home'
                }
			}
        })
        .state('entities.results', <any>{
            parent: 'entities',
            url: '/secret?param1&param2&param3',
            templateUrl: 'app/secret.view.html',
            controller: 'SecretController',
            controllerAs: 'secret',
            requireADLogin: true
        });

	$urlRouterProvider.otherwise('/');
}

```

jasteph: You will not in the documentation http://angular-ui.github.io/ui-router/site/#/api/ui.router.state.$state that the '$stateChangeStart' event supports all the parameters I added.

jasteph:
Can you clarify what you mean by changing this at the "plain js" level? I'd be willing to add these changes if I can understand.

It's my understanding that adal-angular.js handles saving/reloading the redirect page upon return from login with CONSTANTS.STORAGE.START_PAGE. I don't see any further logic for handling this in adal.js.

I'll spend some more time looking at adal.js.

jasteph:
I don't think any further changes need to be made at the plain js level.

weijjia:
sorry for the confusion. adal.js is the plain js level I was referring to. adal.js is saving the login in Constnts.Storage.LOGIN_REQUEST, if there are parameters specified on the login url, we should also preserve those params.

weijjia:
Sorry for the confusion, adal.js is the plain js level I was referring to. adal.js use CONSTANTS.STORAGE.LOGIN_REQUEST to save the login url. I think we also need to preserve the params here.

jasteph:
@weijjia 

I understand now. I have confirmed that the issue does not appear in the adal.js implementation. The plain js implementation uses window.location which successfully captures where the user was attempting to navigate along with the params. Upon return it retrieves the stored window.location and redirects appropriately with the original params.

PR #218, opened 2016-01-19
Added support if iframe and redirect isn't allowed

DariuS231:
There are cases when developing with adaljs where the use of iframes,
redirecting to another page or even changing the the window.location
object is not allowed, for example when developing a google chrome
extension and using the adaljs library.

By adding some extra properties to the initial configuration, the
developer would be able to specify whether the use of iframes and/or
redirect is allowed, as well as a logOutCall function in case the
environment doesn’t allow modifications on the window.location object.

### DIFF
--- a/lib/adal.js
+++ b/lib/adal.js
@@ -134,11 +134,26 @@ AuthenticationContext = function (config) {
     window.callBackMappedToRenewStates = {};
     window.callBacksMappedToRenewStates = {};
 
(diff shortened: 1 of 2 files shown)


tushargupta51:
This library is written primarily for apps running in a regular browser. Hosted browsers and native apps are not supported (am not quite familiar with google extensions, please correct me if these extensions are running in regular browser mode only). In regular browsers, redirection and iframe support are allowed. Currently, we don't intend to extend the support beyond that.  Please let us know if you have any questions. Thanks!

PR #220, opened 2016-01-22
Adding token and logging changes

dmxfee:
- Adding token handling changes
- Changed logging messages
- Updated branch to 1.0.7


weijjia:
did you notice the merge conflicts? Can you fix the merge conflicts? You should do a merge from dev to your branch first. The version is already 1.0.7 in dev.

dmxfee:
Ok, will merge from dev and apply this patch.

lib/adal.js
```diff
@@ -98,6 +98,9 @@ AuthenticationContext = function (config) {
         ERR_MESSAGES: {
             NO_TOKEN: 'User is not authorized'
         },
+        LOADFRAME_TIMEOUT: '15000',
+        LOADFRAME_PENDING: 'pending',
+        LOADFRAME_TIMEOUT_CODE: 'AADSTSTimeout',
```
tushargupta51: AAD uses "AADSTS" prefix for errors originated at AAD. And it might not be an AADSTSTimeout, some other error might cause network issues and kill the current requests. Also, we don't use codes for errors in adaljs, we use `adal:` prefix for angular broadcasts though. We can remove this property.

lib/adal.js
```diff
@@ -316,13 +320,43 @@ AuthenticationContext.prototype._renewIdToken = function (callback) {
...
+ * */
+AuthenticationContext.prototype._loadFrameTimeout = function (urlNavigation, frameName, resource, timeout) {
+    //set iframe session to pending
+    this.verbose('Set loading state to pending for: ' + resource);
+    window.loadframeStatus[resource] = this.CONSTANTS.LOADFRAME_PENDING;
+
+    this._loadFrame(urlNavigation, frameName);
+
+    var timeoutDuration = timeout/1000;
+    var resourceSaved = resource;
```
tushargupta51: Why do we need to save the resource locally? I see we access this value inside the timeout function, is the passed `resource` variable not available inside that method?

lib/adal.js
```diff
@@ -316,13 +320,43 @@ AuthenticationContext.prototype._renewIdToken = function (callback) {
...
+    var self = this;
+
+    setTimeout(function () {
+        if (window.loadframeStatus[resource] === self.CONSTANTS.LOADFRAME_PENDING) {
+            //fail the iframe session if it's in pending state
+            self.verbose('Loading frame has timed out after: ' + timeoutDuration + ' seconds for resource ' + resourceSaved);
+
+            var expectedState = self._activeRenewals[resource];
+            if (expectedState && window.callBackMappedToRenewStates[expectedState]) {
+                window.callBackMappedToRenewStates[expectedState](self.CONSTANTS.LOADFRAME_TIMEOUT_CODE + ' due to timeout', null);
```
tushargupta51: Change the message to 'Adal token renewal operation failed due to timeout'.

lib/adal.js
```diff
@@ -367,7 +401,8 @@ AuthenticationContext.prototype.acquireToken = function (resource, callback) {
 
     // refresh attept with iframe
     //Already renewing for this resource, callback when we get the token.
-    if (this._activeRenewals[resource]) {
+    if (this._activeRenewals[resource] ||
+        (window.parent && window.parent.AuthenticationContext()._activeRenewals[resource])) {
```
tushargupta51: Why did you add the additional checks? The renewal operation always happens in an iframe, I dont think we need to check the window.parent property here.

lib/adal.js
```diff
@@ -558,9 +594,9 @@ AuthenticationContext.prototype.isCallback = function (hash) {
     hash = this._getHash(hash);
     var parameters = this._deserialize(hash);
     return (
-        parameters.hasOwnProperty(this.CONSTANTS.ERROR_DESCRIPTION) ||
-        parameters.hasOwnProperty(this.CONSTANTS.ACCESS_TOKEN) ||
-        parameters.hasOwnProperty(this.CONSTANTS.ID_TOKEN)
+    parameters.hasOwnProperty(this.CONSTANTS.ERROR_DESCRIPTION) ||
+    parameters.hasOwnProperty(this.CONSTANTS.ACCESS_TOKEN) ||
+    parameters.hasOwnProperty(this.CONSTANTS.ID_TOKEN)
     );
```
tushargupta51: Please revert back the indentation.

lib/adal.js
```diff
@@ -684,6 +720,10 @@ AuthenticationContext.prototype.saveTokenFromHash = function (requestInfo) {
                 // save token with related resource
                 this._saveItem(this.CONSTANTS.STORAGE.ACCESS_TOKEN_KEY + resource, requestInfo.parameters[this.CONSTANTS.ACCESS_TOKEN]);
                 this._saveItem(this.CONSTANTS.STORAGE.EXPIRATION_KEY + resource, this._expiresIn(requestInfo.parameters[this.CONSTANTS.EXPIRES_IN]));
+                if (window.parent && window.parent.loadframeStatus) {
```
tushargupta51: We need this check when aad returns an error in the if block above line #700. Since, we received the response, even if its error, we should delete the loadframestatus so that its no longer pending.

lib/adal.js
```diff
@@ -684,6 +720,10 @@ AuthenticationContext.prototype.saveTokenFromHash = function (requestInfo) {
...
+                    delete window.parent.loadframeStatus[resource];
+                }
+                this.verbose('token sucessfully updated in local storage. removing iframe session tracking for ' + resource);
```
tushargupta51: nit: capitalize the first letter. "Token", "Rremoving"

lib/adal.js
```diff
@@ -705,6 +745,10 @@ AuthenticationContext.prototype.saveTokenFromHash = function (requestInfo) {
                         }
                         this._saveItem(this.CONSTANTS.STORAGE.ACCESS_TOKEN_KEY + resource, requestInfo.parameters[this.CONSTANTS.ID_TOKEN]);
                         this._saveItem(this.CONSTANTS.STORAGE.EXPIRATION_KEY + resource, this._user.profile.exp);
+                        if (window.parent && window.parent.loadframeStatus) {
+                            delete window.parent.loadframeStatus[resource];
+                        }
+                        this.verbose('token sucessfully updated in local storage. removing iframe session tracking for ' + resource);
```
tushargupta51: same nit as above

lib/adal.js
```diff
@@ -770,6 +814,7 @@ AuthenticationContext.prototype.handleWindowCallback = function () {
             // iframe call but same single page
             this.verbose('Window is in iframe');
             callback = window.parent.callBackMappedToRenewStates[requestInfo.stateResponse];
+            this.verbose('iframe came back, deleting iframe session tracking for ' + requestInfo.stateResponse);
```
tushargupta51: same nit as above

lib/adal.js
```diff
@@ -316,13 +320,43 @@ AuthenticationContext.prototype._renewIdToken = function (callback) {
...
 
 AuthenticationContext.prototype._urlContainsQueryStringParameter = function (name, url) {
     // regex to detect pattern of a ? or & followed by the name parameter and an equals character
     var regex = new RegExp("[\\?&]" + name + "=");
     return regex.test(url);
+};
+
+/* Calling _loadFrame but with a timeout to signal failure in loadframeStatus. Frames are left
+ * hanging when network errors occur.
+ * https://github.com/AzureAD/azure-activedirectory-library-for-js/issues/250.
```
tushargupta51: Please remove the reference to the issue, we do not add those.

tushargupta51:
@dmxfee Thanks for your contribution! I know it has been a while since you sent the PR, sorry for the delay. We want to fix this issue in next release. If you are still interested in getting this PR merged, I have added some comments in the PR. If you do not have the bandwidth to work on this, that's fine, let us know, I will create another PR inspired from your changes and add you as a contributor. We need some unit tests also to make sure the new behavior works fine.

lib/adal.js
dmxfee: Good find

lib/adal.js
```diff
@@ -378,6 +412,7 @@ AuthenticationContext.prototype.acquireToken = function (resource, callback) {
             this.verbose('renewing idtoken');
             this._renewIdToken(callback);
         } else {
+            this.verbose('renewing token ' + resource);
```
tushargupta51: I don't think we need this logging statement here. We have one in the renewToken method: this.info('renewToken is called for resource:' + resource)

lib/adal.js
```diff
@@ -672,18 +708,18 @@ AuthenticationContext.prototype.saveTokenFromHash = function (requestInfo) {
...
                 this.info('Fragment has access token');
-                resource = this._getResourceFromState(requestInfo.stateResponse);
                 if (!this._hasResource(resource)) {
                     keys = this._getItem(this.CONSTANTS.STORAGE.TOKEN_KEYS) || '';
                     this._saveItem(this.CONSTANTS.STORAGE.TOKEN_KEYS, keys + resource + this.CONSTANTS.RESOURCE_DELIMETER);
                 }
                 // save token with related resource
                 this._saveItem(this.CONSTANTS.STORAGE.ACCESS_TOKEN_KEY + resource, requestInfo.parameters[this.CONSTANTS.ACCESS_TOKEN]);
                 this._saveItem(this.CONSTANTS.STORAGE.EXPIRATION_KEY + resource, this._expiresIn(requestInfo.parameters[this.CONSTANTS.EXPIRES_IN]));
+                this.verbose('Token successfully updated in local storage.');
```
tushargupta51: Can you update this to "access token updated in storage."? The storage used could be session storage.

lib/adal.js
```diff
@@ -705,6 +741,7 @@ AuthenticationContext.prototype.saveTokenFromHash = function (requestInfo) {
                         }
                         this._saveItem(this.CONSTANTS.STORAGE.ACCESS_TOKEN_KEY + resource, requestInfo.parameters[this.CONSTANTS.ID_TOKEN]);
                         this._saveItem(this.CONSTANTS.STORAGE.EXPIRATION_KEY + resource, this._user.profile.exp);
+                        this.verbose('token successfully updated in local storage.');
```
tushargupta51: Same as above. Please update the statement to "access token updated in storage."?

lib/adal.js
```diff
@@ -713,6 +750,11 @@ AuthenticationContext.prototype.saveTokenFromHash = function (requestInfo) {
             this._saveItem(this.CONSTANTS.STORAGE.ERROR_DESCRIPTION, 'Invalid_state. state: ' + requestInfo.stateResponse);
         }
     }
+
+    if (window.parent && window.parent.loadframeStatus && resource) {
```
tushargupta51: For detecting iframe, the check should be `window !== window.parent`. I know we have a similar check in handleWindowCallback method, that needs to be updated too.

lib/adal.js
tushargupta51: I don't think we need this logging statement here.

tushargupta51:
@dmxfee As per our discussion, closing this PR. The changes are part of #317 PR.

PR #225, opened 2016-02-01
prevent login loop when using directives that make http calls to prot…

rdmchenry:
…ected resource before adal login is initiated

A login loop condition occurs when using directives that
make HTTP calls to the ADAL protected API. The HTTP
request from the directive is made prior to ADAL initialization
so the call does not get cancelled in the request interceptor
as a login is not in progress yet. 

The ADAL login then starts
and completes caching the token. Now the directive HTTP
request returns to the response interceptor with a 401. This
triggers a token cache clearing for the resource and another
login begins with the next HTTP request in the app.


weijjia:
@rdmchenry As we talked, let's hold off this PR for now.

Issue #195, opened 2015-12-07
Token renew only works if endpoints are passed to config, even thought it's marked optional


Valchris:
I believe the offending code lies in ProtectedResourceInteceptor, the else if condition should actually be:
```javascript
} else if (authService.config && (!authService.config.endpoints || isEndpoint)) {
```
Snippet below:
```javascript
if (authService.loginInProgress()) {
    authService.info('login already start.');
    return $q.reject();
} else if (authService.config && isEndpoint) {
    // external endpoints
    // delayed request to return after iframe completes
    var delayedRequest = $q.defer();
    authService.acquireToken(resource).then(function (token) {
    authService.verbose('Token is avaliable');
    config.headers.Authorization = 'Bearer ' + token;
    delayedRequest.resolve(config);
}, function (err) {
    delayedRequest.reject(err);
});
```

AndreSayreMSFT:
@weijjia We seem to be hitting this, or at least a related problem, internally!

The existence of this issue prevents the Authorization token from being attached to the first of our outgoing $http calls, thus causing all sorts of badness for us.

CC: @yanchenw

tushargupta51:
Fixed in #240.

PR #240, opened 2016-03-02
Fix for the issue #195 - Token renew only works if endpoints are pass…

sarmut:
This fix addresses the issue #195.  When the endpoints are not provided, the renew call was not issued.  This change set fixes that check.

### DIFF
--- a/lib/adal-angular.js
+++ b/lib/adal-angular.js
@@ -300,27 +300,17 @@ if (typeof module !== 'undefined' && module.exports) {
                         }
 
                         var tokenStored = authService.getCachedToken(resource);
-                        var isEndpoint = false;
                         if (tokenStored) {
                             authService.info('Token is avaliable for this url ' + config.url);
                             // check endpoint mapping if provided
                             config.headers.Authorization = 'Bearer ' + tokenStored;
                             return config;
                         } else {
-
-                            if (authService.config) {
-                                for (var endpointUrl in authService.config.endpoints) {
-                                    if (config.url.indexOf(endpointUrl) > -1) {
-                                        isEndpoint = true;
-                                    }
-                                }
-                            }
-
                             // Cancel request if login is starting
                             if (authService.loginInProgress()) {
                                 authService.info('login already start.');
                                 return $q.reject();
-                            } else if (authService.config && (!authService.config.endpoints || isEndpoint)) {
+                            } else {
                                 // external endpoints
                                 // delayed request to return after iframe completes
                                 var delayedRequest = $q.defer();
--- a/lib/adal-angular.js
+++ b/lib/adal-angular.js
@@ -320,7 +320,7 @@ if (typeof module !== 'undefined' && module.exports) {
                             if (authService.loginInProgress()) {
                                 authService.info('login already start.');
                                 return $q.reject();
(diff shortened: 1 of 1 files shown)


lib/adal-angular.js
```diff
@@ -320,7 +320,7 @@ if (typeof module !== 'undefined' && module.exports) {
                             if (authService.loginInProgress()) {
                                 authService.info('login already start.');
                                 return $q.reject();
-                            } else if (authService.config && isEndpoint) {
+                            } else if (authService.config && (!authService.config.endpoints || isEndpoint)) {
```
weijjia: I'm a little bit concerned doing this way. What we really want to do is checking if it's in endpoint list or from app backend. But we already return if it's not from app backend or in endpoint list when checking the resource. Let me give a little bit more thought on this and get back to you.

tushargupta51: @sarmut We already check at line 297 if the given uri matches any configured endpoints or app's client_id or not. We are again checking it in line 311-317 which is not needed. I think we can get of lines 311-317 and change the `else if` check to just `else` since we want to make that request for both external endpoints and app's backend. Also, please update the comment below that check.

sarmut: @tushargupta51 I agree to your point.  The check on line 311-317 is duplicate.  But I have a question.. if line#297 gives a non-null 'resource', how the same duplicate check failed to update the flag isEndpoint as true?  Does this mean, the check on #298 is wrong too ?

tushargupta51: @sarmut `getResourceForEndpoint` at line 297 does an additional check matching the uri against app's client id. In that case, isEndpoint would be false, because `client_id` is not part of config.endpoints. app's client_is is the default endpoint.

sarmut: @tushargupta51 I don't see the client id matching.. Noted below the lines I'm looking.  Can you please explain, how the first one is including the 'client id' check?

In adal.js, line#769-776

AuthenticationContext.prototype.getResourceForEndpoint = function (endpoint) {
    if (this.config && this.config.endpoints) {
        for (var configEndpoint in this.config.endpoints) {
            // configEndpoint is like /api/Todo requested endpoint can be /api/Todo/1
            if (endpoint.indexOf(configEndpoint) > -1) {
                return this.config.endpoints[configEndpoint];
            }
        }
    }

in adal-angular.js, line#311-317

if (authService.config) {
   for (var endpointUrl in authService.config.endpoints) {
   if (config.url.indexOf(endpointUrl) > -1) {
      isEndpoint = true;
   }
 }
}

sarmut: Got it.. I can submit the updated code as a PR.  @weijjia, can you please let me know what you think?  Do you have any other solution?

weijjia: @sarmut I've already talked with @tushargupta51, and we're in agreement with this change.

sarmut:
@weijjia @tushargupta51 submitted the change as discussed.  Please approve and merge it with the master branch.

tushargupta51:
Thanks @sarmut for your contribution! Changes look good, will merge them.

sarmut:
Thank you @tushargupta51 .  I very much appreciate it :)

PR #242, opened 2016-03-02
Add support for AMD and window pinning

Mike-Bell:
Just followed the suggestions here:

http://ifandelse.com/its-not-hard-making-your-library-support-amd-and-commonjs/

This allows people consuming the adal.js library directly without the angular wrapper (e.g. me) to have more flexible options for importing the library.


Mike-Bell:
Looks like most people are PRing against dev, and that I'm a bit behind. I'll refetch and reopen against dev.

PR #243, opened 2016-03-02
Add support for AMD and window pinning

Mike-Bell:
Just followed the suggestions here:

http://ifandelse.com/its-not-hard-making-your-library-support-amd-and-commonjs/

This allows people consuming the adal.js library directly without the angular wrapper (e.g. me) to have more flexible options for importing the library.


rohitnarula7176:
Hi Mike,
Thank you for the suggestion. Please make the following change and then we can merge your pull request:
1. Do not submit the adal.min file as that file is auto-generated.
2. Let the following code be there in the adal.js:
 if (typeof module !== 'undefined' && module.exports) {
        module.exports.inject = function (conf) {
            return new AuthenticationContext(conf);
        };
    }
3. Make a similar change in adal-angular.js to define adal as a dependency as mentioned below:
(function (root, factory) {
    if(typeof define === "function" && define.amd) {
        define(["./adal"], factory);
    } else if(typeof module === "object" && module.exports) {
        module.exports = factory(require("./adal"));
    } else {
        root.AdalModule = factory(root.AuthenticationContext);
    }
}(this, function(AuthenticationContext) {

4. At the end of adal-angular.js code. Add the following line to return the adal module:
return AdalModule;

tushargupta51:
@Mike-Bell Thanks for submitting the PR. We recently made a change (its in dev branch) where we use IIFE pattern and expose AuthenticationContext on module. We tested that with "browserify", "npm", "webpack" and simple "script" tags and it worked fine with all these different ways of consuming adal. AdalJS is strictly a front end library and is not meant to be run on server side. 

Try consuming the adal.js from dev branch and let us know if you are still having issues with the current code.

Mike-Bell:
`IIFE` is a step in the right direction, but it looks like you're still missing AMD support (for Require.js in particular). It usually starts like this `if(typeof define === "function" && define.amd)`. For good measure, I'd also support pinning AuthenticationContext to the global scope, since starting your script by defining a `var` and hoping that others will be able to use it can be problematic. If I get a chance I'll update this PR. It should diff more cleanly now that we're in an IIFE.

tushargupta51:
@Mike-Bell The problem with pinning AuthenticationContext to global scope, as many users have reported, is it gets wrapped inside a module when using loaders like webpack and then its not available globally unless you explicitly export in `require` statement. Maybe, it was happening because we didn't have the IIFE pattern earlier.

@rohitnarula7176 Can you test whether adal is consumable with AMD/Require.js?

PR #252, opened 2016-03-23
js: Add optional "version" parameter to config

rstropek:
AAD B2C requires "v2.0" in the authorization endpoint. There has not
been a way to make ADAL JS generate a login URL like
".../oauth2/v2.0/authorize...".

This commit adds an optional parameter "version" to the config object.
If it is set, it is injected in the login URL. If it is not set,
the behavior does not change to previous versions.

This is not full support for AAD B2C. However, it would be very
helpful until full support for AAD B2C comes for ADAL JS.

### DIFF
--- a/lib/adal.js
+++ b/lib/adal.js
@@ -846,8 +846,13 @@ AuthenticationContext.prototype._getNavigateUrl = function (responseType, resour
     if (this.config.instance) {
         this.instance = this.config.instance;
     }
+	    
+    var version = '';
+    if (this.config.version) {
+        version = this.config.version + '/';
+    }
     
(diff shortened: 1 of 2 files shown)


tushargupta51:
@rstropek Thanks for your contribution. Adding an additional value on config does not seem the best way to handle this. You can specify `adal.config.displayCall(url)` extensibility point (https://github.com/AzureAD/azure-activedirectory-library-for-js/blob/master/lib/adal.js#L183) and modify the url by adding v2.0 before sending the request.

tushargupta51:
@rstropek You can use experimental version of adal to authenticate against v2.0 but keep in mind that is an experimental version. Also, just adding the version wont work well, there are other changes needed to be made for adal to work against v2.0. Closing this PR as we are adding support for v2.0 in experimental branch.

rstropek:
Thank you. I will take a look at the experimental version.

Issue #247, opened 2016-03-17
Login fails with Invalid_state when running on Edge and when my portal and AAD are in a different security zones


ischrei:
Hi,

I've found that sessionStorage isn't preserved in Edge, when I'm trying to login using AAD to my portal, and when the sites are in different security zones (i.e. AAD login is in the trusted zone and mine is under Internet zone).

Since the storage isn't preserved, the login state value it gone, so login fails with Invalid_state error.
I know that according to the documentation my portal and AAD have to be in the same zone (due to other reasons as well), but I would like let the user know the exact issue and how to resolve it.
However, when this login fails, I don't get any adal:loginFailure event.
1. Why the event isn't sent?
2. Is there a recommendation how to identify the portal and AAD are in different security zones?

Thanks!
Ido.

tushargupta51:
@ischrei I agree that we should broadcast a loginfailure event in case of state no found, but we cannot identify in adal that the state was not saved because of different security zones. The event will have description like "state not found" or "state not match".

ischrei:
I agree this particular error can be caused by other root causes as well. However, since different security zones is a blocker to the ADAL framework, I believe you should help your users identify this case somehow, so their customers will be notified correctly.

Do you have an ETA for the event fix?

Thanks a lot in advance!

tushargupta51:
@ischrei do you have the zoning issue on other browsers too? Chrome/Firefox?

ischrei:
Only om IE & Edge

Issue #212, opened 2016-01-04
Change npm licenses to license and use spdx syntax


brandwe:
Basically you don’t specify “licenses” anymore but “license” with a string that corresponds to a supported thing: https://spdx.org/licenses/. Details here: 
Should use for all npms: https://docs.npmjs.com/files/package.json

wakurth:
bump.

Issue #180, opened 2015-10-22
add lib version in the log


weijjia:
add lib version in the log entry

Issue #111, opened 2015-04-20
redirecting to a protected resource throws route change error


bigerock:
when you use requireADLogin: true on a route, and redirect to it while not authenticated, it is throwing a route change error and then proceeds to properly redirect to the microsoft login page.

although it is ultimately working, is is throwing the error. it's as if it is not rejecting the route properly. is there anything i can do to prevent this error?

omercs:
What is this route change error? Can you share the details?

bigerock:
yes, i will try. it happens when the user is logged out, and clicks a link that is protected by requireADLogin. 

here's the error:

Error: [$compile:tpload] Failed to load template: /app/dashboard/dashboard.html
http://errors.angularjs.org/1.3.0/$compile/tpload?p0=%2Fapp%2Fdashboard%2Fdashboard.html

and here's the trace:

  at REGEX_STRING_REGEXP (angular.js:80)
    at handleError (angular.js:15773)
    at processQueue (angular.js:12914)
    at angular.js:12930
    at Scope.$get.Scope.$eval (angular.js:14123)
    at Scope.$get.Scope.$digest (angular.js:13939)
    at Scope.$get.Scope.$apply (angular.js:14227)
    at HTMLHtmlElement.<anonymous> (angular.js:11131)
    at HTMLHtmlElement.n.event.dispatch (jquery.min.js:3)
    at HTMLHtmlElement.n.event.add.r.handle (jquery.min.js:3)

Then, it does continue on to the microsoft login.

'app/dashboard/dashboard.html' is the page that is protected and the page we are trying to go to.

Please let me know if you need further details.

omercs:
Interceptor in adal-angular.js cancels the request when login is started for this route.

bigerock:
correct, but this throws a script error before performing the redirect. is there any way to catch or suppress this error?

tushargupta51:
@bigerock we will add an even that you can listen to whenever a request is rejected.

PR #253, opened 2016-03-26
Bug fixes

### LINKED ISSUE #249: LOGIN_ERROR isn't set, as getRequestInfo doesn't set the requestType correctly
When there is an invalid_state error, the getRequestInfo  fails to set the requestType as the the state doesn't match.
The requestType valu is set only when there is a match, and is taken from the sessionStorage.

You should get the request type from the returned state parameter in the hash.

tushargupta51:
#249 #247 : broadcasting 'adal:stateMismatch' error whenever invalid state is received.
#212 : updating the 'license' field in package.json
#180 : added adal library version in log messages, each message will contain the version, we can consider adding only once but it does not matter so much, method returns a constant and its not going to clutter messages.
#111 : whenever interceptor rejects a request because a login is in progress, adal broadcase 'adal:errorResponse' with the rejection message, this event will be raised if some other error happened (except 401 for which there is a different event) when interceptor processing the request.
#76 : we maintain a flag adal._renewActive, if the flag is set, we do not initiate another login request while a renew request is in progress.

Other then the above bugs:
1. Removed state.idToken and failed.renew storage constants since they are redundant and are not needed. we do not try to acquire token again if the previous request failed for some reason, this does not look correct behavior to me. In case of failure, we do send the error to the registered callback.
2. Added tests
3. Cleanup

@polita @brentschmaltz @weijjia


lib/adal.js
```diff
@@ -176,11 +171,10 @@ AuthenticationContext.prototype.login = function () {
     this._saveItem(this.CONSTANTS.STORAGE.LOGIN_ERROR, '');
     this._saveItem(this.CONSTANTS.STORAGE.STATE_LOGIN, expectedState);
     this._saveItem(this.CONSTANTS.STORAGE.NONCE_IDTOKEN, this._idTokenNonce);
-    this._saveItem(this.CONSTANTS.STORAGE.FAILED_RENEW, '');
     this._saveItem(this.CONSTANTS.STORAGE.ERROR, '');
     this._saveItem(this.CONSTANTS.STORAGE.ERROR_DESCRIPTION, '');
-    
-    
+
+
```
weijjia: nit: extra line

lib/adal.js
```diff
@@ -1187,7 +1141,7 @@ AuthenticationContext.prototype._cloneConfig = function (obj) {
     return copy;
 };
 
-AuthenticationContext.prototype._addClientId = function () {
+AuthenticationContext.prototype._addXClientParameters = function () {
```
weijjia: the name is not so clear, we're actually adding lib version.

package.json
```diff
@@ -5,15 +5,10 @@
...
-      "url": "http://www.apache.org/licenses/LICENSE-2.0"
-    }
-  ],
+  "license": "Apache-2.0",
   "repository": {
     "type": "git",
-    "url": ""
+    "url": "https://github.com/AzureAD/azure-activedirectory-library-for-js.git"
   },
   "version": "1.0.8",
```
weijjia: this version should also be updated if you're going to change the version to 1.0.9

lib/adal.js
```diff
@@ -709,28 +672,28 @@ AuthenticationContext.prototype.saveTokenFromHash = function (requestInfo) {
...
-            
+
             var keys, resource;
-            
+
             if (requestInfo.parameters.hasOwnProperty(this.CONSTANTS.ACCESS_TOKEN)) {
                 this.info('Fragment has access token');
                 // default resource
                 resource = this.config.loginResource;
+                if (requestInfo.requestType === this.REQUEST_TYPE.RENEW_TOKEN) {
```
weijjia: I'm thinking why we have this check here? if the fragment has access token, it can only be renew token request. If so, we should always be able tell the resource based on state response, then why setting the default resource?

weijjia:
For the broadcasting error message, as long as everybody agrees on it, then we should be good. If we start broadcasting those errors at the angular level, developer would rely on them, then we need to be cautions when changing them.

lib/adal.js
tushargupta51: You are correct. I don't think resource = this.config.loginResource is needed.

lib/adal.js
tushargupta51: we are adding x-client-SKU and x-client-ver parameters, thats why named it addXClientParameters.

weijjia: oh, sorry, i miss the x-client-sku. XClientParmeters is still not so clear to me, I have to look into the method for what it is adding.

package.json
tushargupta51: thanks for this, updated the version at other places too.

lib/adal.js
tushargupta51: Since both of these values represent information about the library, sku==platform, ver==version, another names could be _addLibraryMetadata or _addMetadata.

weijjia: i vote for  _addLibraryMetadata

lib/adal-angular.js
```diff
@@ -89,6 +89,7 @@ if (typeof module !== 'undefined' && module.exports) {
                         }
 
                         if (requestInfo.requestType !== _adal.REQUEST_TYPE.LOGIN) {
+                            _adal._renewActive = false;
```
ischrei: Sorry I'm putting the comment in a wrong line - I think we should add a broadcast line in line 111 for errors, as the returned hash may contain the errro_description from AAD.:
if (requestInfo.parameters['access_token']) {
                                        _adal.callback(_adal._getItem(_adal.CONSTANTS.STORAGE.ERROR_DESCRIPTION), requestInfo.parameters['access_token']);
                                        return;
                                    } else if (requestInfo.parameters['id_token']) {
                                        _adal.callback(_adal._getItem(_adal.CONSTANTS.STORAGE.ERROR_DESCRIPTION), requestInfo.parameters['id_token']);
                                        return;
                                    } else {
                                        $rootScope.$broadcast('adal:loginFailure', _adal._getItem(_adal.CONSTANTS.STORAGE.ERROR_DESCRIPTION));
                                    }

package.json
```diff
@@ -5,15 +5,10 @@
     "email": "msopentech@microsoft.com",
     "url": "https://github.com/AzureAD/azure-activedirectory-library-for-js"
```
polita: Should this still be published as MSOT, or should we switch it to Microsoft?

package.json
polita: Please make sure it's updated in Readme.md

---
In reply to: [57949952](https://github.com/AzureAD/azure-activedirectory-library-for-js/pull/253#discussion_r57949952) [](ancestors = 57949952)

tushargupta51: I had a local commit that updates other files, just pushed it out to github.

polita: Ah, yes, now I see the new iteration. Thanks.

---
In reply to: [58117639](https://github.com/AzureAD/azure-activedirectory-library-for-js/pull/253#discussion_r58117639) [](ancestors = 58117639)

README.md
```diff
@@ -7,7 +7,7 @@ This library is optimized for working together with AngularJS.
 
 ## The Library
 
-This is a GA released version. The current version is **1.0.8**.
+This is a GA released version. The current version is **1.0.9**.
```
polita: Do the paths below and the CDN version need to change for the new version?

lib/adal-angular.js
tushargupta51: @ischrei Thanks for catching this! I agree that we need to report the error somehow but the loginFailure event does not seem to be the right one to be broadcasted. This error is likely to occur at renew time rather than login. We are on a tight schedule to release 1.0.9, I will fix this up in next release. If this is not yet reported, please feel free to open an issue.

tushargupta51:
@weijjia I talked with @brentschmaltz and @polita regarding the new broadcast events and we agreed on adding those events.

weijjia:
@tushargupta51 Good to know, thanks for the info. :)

package.json
tushargupta51: This should be switched to Microsoft. how about nugetaad@microsoft.com? Thats what we add in our wilson nuget packages. Now that we are using this email address for npm, makes sense to use that email address here too.

README.md
```diff
@@ -7,17 +7,17 @@ This library is optimized for working together with AngularJS.
...
     <!-- Latest compiled and minified JavaScript -->
-    <script src="https://secure.aadcdn.microsoftonline-p.com/lib/1.0.8/js/adal.min.js"></script>
-    <script src="https://secure.aadcdn.microsoftonline-p.com/lib/1.0.8/js/adal-angular.min.js"></script>
+    <script src="https://secure.aadcdn.microsoftonline-p.com/lib/1.0.9/js/adal.min.js"></script>
+    <script src="https://secure.aadcdn.microsoftonline-p.com/lib/1.0.9/js/adal-angular.min.js"></script>
 
-CDN will be updated to latest version 1.0.8.
+CDN will be updated to latest version 1.0.9.
 
 Via Bower:
```
polita: We should have a new section here for "Via NPM"

tushargupta51: Added.

PR #255, opened 2016-04-01
set the state when renew id token

hejiantao5:
if not set this value, line 648 will never matched even it is renew id token actually


hejiantao5:
OK, I saw you patch '|| requestInfo.parameters[this.CONSTANTS.ID_TOKEN]' in line 830, my local version is out of date. But in my scenario the ACCESS_TOKEN is not a JWT token, so I suggest to use ' requestInfo.parameters[this.CONSTANTS.ID_TOKEN] || requestInfo.parameters[this.CONSTANTS.ACCESS_TOKEN]'

hejiantao5:
Ok, thank you @tushargupta51

PR #258, opened 2016-04-05
$timeout needed to apply scope changes

nillis:


lib/adal-angular.js
```diff
@@ -155,7 +155,9 @@ if (typeof module !== 'undefined' && module.exports) {
                                     $rootScope.$broadcast('adal:loginFailure', 'auto renew failure');
                                 } else {
                                     if (tokenOut) {
-                                        _oauthData.isAuthenticated = true;
+                                        $timeout(function () {
+                                            _oauthData.isAuthenticated = true;
```
tushargupta51: @nillis This isn't a scope change. In other places, we are making changes to rootscope, and hence timeout was needed.

tushargupta51:
@nillis Thanks for taking out time to submit this PR. As I mentioned in my comment, we are not making any scope changes in that else block, so a timeout to apply the changes is not needed there. However, if you think otherwise, I would recommend opening an issue to discuss the problem. And then if we agree on the issue, we can have a PR with the fix.

Issue #266, opened 2016-04-13
Authentication Error on non-requireADLogin routes


Blackbaud-ScottVossen:
I have a fairly simple angular application using ui.router where I have a homepage which you don't have to be logged in to view, and other pages you need to be logged in to view:

        $urlRouterProvider.otherwise('/');

        $stateProvider
            .state('home', {
                url: '/',
                templateUrl: 'app/home/home.html',
                controller: 'homeController as ctrl'
            })

            .state('dashboard', {
                url: '/dashboard',
                requireADLogin: true,
                views: {
                    '' : {
                        templateUrl: 'app/dashboard/dashboard.html',
                        controller: 'dashboardController as ctrl'
                    },
                    'fundSummaryTile@dashboard': {
                        templateUrl: 'app/dashboard/fund-summary-tile.html',
                        controller: 'fundSummaryTileController as ctrl'
                    }
                }
            })

            .state('fund', {
                url: '/fund/:id',
                requireADLogin: true,
                views: {
                    '': {
                        controller: 'fundDetailController as ctrl',
                        templateUrl: 'app/fund/fund-detail.html'
                    }
                }
            });

This worked just fine in version 1.0.8, but with the update to 1.0.9, this is now throwing a **stateChangeError** during the transition. The error shows the correct toState, but results in a **"User login is required" error message**. 

I switched my version back to 1.0.8 and things were working perfectly again. If there's something new I need to do to get this working in 1.0.9, please let me know.

JoshDavisMP:
Commit 2895778dbe1b399ee47f5338786592728a975bfb to fix #195 removed the check on endpoints so now **all** service requests require authentication.

gpit2286:
Currently, there is the option to `requireADLogin: false` in the state config that does not seem to work either.  I see that fixes you made in #275 but I wonder if the anonymous endpoint solution is the way to go.  Does that mean that we will no longer be able to specify the `requireADLogin` option as it would be obsolete?

Blackbaud-ScottVossen:
@JoshDavisMP, so the functionality that allows someone to have a homepage that doesn't require you to be logged in is being removed? I see #275 addresses anonymous endpoints, but this doesn't seem to allow me to have a user visit one of my pages without being authenticated.

If there's a new way to accomplish this, please fill me in or point me to the correct documentation!

Please don't remove my requireADLogin!

JoshDavisMP:
[...]

Issue #264, opened 2016-04-11
1.0.9 page endless refresh


sjaanus:
Configuration:

```
$routeProvider
			.when('/', { redirectTo: '/attendancereport', requireADLogin: true })
			.when('/admin', { templateUrl: '/admin/home', requireADLogin: true })
			.otherwise({ redirectTo: '/' });

```

```

		adalProvider.init({
			instance: 'https://login.microsoftonline.com/',
			tenant: 'mytenant',
			clientId: 'clientidb',
			extraQueryParameter: 'nux=1',
		}, $httpProvider);
```

When I open the root page , it just keeps refreshing everything. Any suggestions what is wrong?

This happens for example, when page has been not been used for 50 minutes, the session storage also changed and it keeps refreshing itself forever. I notice that **adal.access.token.key** goes missing.

http://imgur.com/0zXK9Yq

tushargupta51:
@sjaanus Can you share a log trace? You can add this to turn on console logging:

Logging = {
    level: 3,
    log: function (message) {
        console.log(message);
    }
};

ronnbot:
I experience this when I test on a Windows 10 environment (Edge or Chrome). There are no error though.

PavelPikat:
You need to change your "otherwise" state to a route that is not protected by AD.
You can also try adding these lines of code that I posted earlier for a similar issue. This will help to get rid of scope digesting:
https://github.com/AzureAD/azure-activedirectory-library-for-js/issues/145#issuecomment-205687778
and
https://github.com/AzureAD/azure-activedirectory-library-for-js/issues/145#issuecomment-205166820

krushnachandra:
// set default routes 
        $urlRouterProvider.when('', '/index');
        $urlRouterProvider.when('/', '/index');
        $urlRouterProvider.otherwise('/index');


in controller of index

 if ($rootScope.userInfo.isAuthenticated) {
                $state.go('routeloader');
            }
            else {
                $state.go('index');
            }


============
  $stateProvider
        .state('routeloader', {
            url: '/temp',
            templateUrl: 'app/setup/routeloader.tmpl.html',
            controller: 'routeLoaderCtrl',
            controllerAs: 'vm',
           //dont use adlogin:true it is temporary state for some logic for application like check claims                   and all in the meantime show a loader to user
           //if satisfied the login then move to dashboard otherwise redirect to home 
        });

sjaanus:
[...]

Issue #261, opened 2016-04-06
Interceptor problem on local requests coming from $templateCache


clembou:
adding and configuring the library breaks my app, even if no authentication is requested anywhere.

The interceptor intercepts `get` requests coming from `$templateCache` to angular template relative urls, treats them as local resources, and expect to be able to find a token for these, see [here](https://github.com/AzureAD/azure-activedirectory-library-for-js/blob/master/lib/adal.js#L744).

As a result, a call to acquire a new token using the app `clientid` is generated [here](https://github.com/AzureAD/azure-activedirectory-library-for-js/blob/master/lib/adal-angular.js#L322) if no token is found in the cache, but this fails because the user is not logged in and authentication has not been requested, see  [here](https://github.com/AzureAD/azure-activedirectory-library-for-js/blob/master/lib/adal.js#L362)

Log:
```
Wed, 06 Apr 2016 17:39:20 GMT:843fe150-a8d0-4f30-bd2d-a6a6bd3f74c3-1.0.9-WARNING: User login is required
Wed, 06 Apr 2016 17:39:20 GMT:843fe150-a8d0-4f30-bd2d-a6a6bd3f74c3-1.0.9-ERROR: Error when acquiring token for resource: <CLIENTID>
stack:
undefined
Wed, 06 Apr 2016 17:39:20 GMT:843fe150-a8d0-4f30-bd2d-a6a6bd3f74c3-1.0.9-INFO: Getting error in the response
Error: [$compile:tpload] Failed to load template: scripts/components/myComponent.html (HTTP status: undefined undefined)
http://errors.angularjs.org/1.5.2/$compile/tpload?p0=scripts%2Fcomponents%2FmyComponent.html&p1=undefined&p2=undefined
    at angular.js:68
    at handleError (angular.js:18640)
    at processQueue (angular.js:15627)
    at angular.js:15643
    at Scope.$eval (angular.js:16895)
    at Scope.$digest (angular.js:16711)
    at Scope.$apply (angular.js:17003)
    at bootstrapApply (angular.js:1713)
    at Object.invoke (angular.js:4625)
    at doBootstrap (angular.js:1711)
```
I think this stackoverflow issue refers to the same problem:
http://stackoverflow.com/questions/35650241/adal-angular-js-error-failed-to-load-template

[...]

Issue #260, opened 2016-04-06
Callback isn't called in locationChangeHandler when there is an error acvquiring the token


ischrei:
existing code:
if (requestInfo.parameters['access_token']) {
                                        _adal.callback(_adal._getItem(_adal.CONSTANTS.STORAGE.ERROR_DESCRIPTION), requestInfo.parameters['access_token']);
                                        return;
                                    } else if (requestInfo.parameters['id_token']) {
                                        _adal.callback(_adal._getItem(_adal.CONSTANTS.STORAGE.ERROR_DESCRIPTION), requestInfo.parameters['id_token']);
                                        return;
                                    }

------------

Suggested fix:
if (requestInfo.parameters['access_token']) {
                                        _adal.callback(_adal._getItem(_adal.CONSTANTS.STORAGE.ERROR_DESCRIPTION), requestInfo.parameters['access_token']);
                                        return;
                                    } else if (requestInfo.parameters['id_token']) {
                                        _adal.callback(_adal._getItem(_adal.CONSTANTS.STORAGE.ERROR_DESCRIPTION), requestInfo.parameters['id_token']);
                                        return;
                                    }
**else {
                                        $rootScope.$broadcast('adal:loginFailure', _adal._getItem(_adal.CONSTANTS.STORAGE.ERROR_DESCRIPTION));
                                        _adal.callback(_adal._getItem(_adal.CONSTANTS.STORAGE.ERROR_DESCRIPTION), null);
                                    }**

(not sure whether we need to "return" in this case).

Thanks,
Ido.

ischrei:
Another issue:
when there is an error acquiring the token, a false 'adal:loginSuccess' is sent.
The locationChangeHandler method is called more than once for this single redirection, and in one of the calls, the 'adal.callback' is null.

tushargupta51:
Thanks @ischrei . I will get it in the next release.

tushargupta51:
1.0.10 fixes this. Please let me know, if you still see issues.

PR #274, opened 2016-04-28
fixes refresh token loop, add anonymousEndpoints extensibility.

tushargupta51:
#266 #261 #127 : These issues are caused because we secure all the requests send to app's backend. Now, I added an extensibility point - anonymousEndpoints that users can specify in init configuration. Adal will not try to attach tokens to requests send to these endpoints.

#264 #202 #216 : These issues are caused when access token is expired and AAD cookie is not good. So, when adal tries to renew the token, AAD sends an error back to ADAL, these forces adal to go in a loop. Now, I set a `_renewFailed` flag to true if this happens and avoid subsequent token renewal requests when we received an error from AAD.

#260: Earlier, when AAD sends error in response, we were not calling the callback registered by the user at the time of calling the `acquireToken` method. The callback is called with the error.

### DIFF
--- a/lib/adal-angular.js
+++ b/lib/adal-angular.js
@@ -1,5 +1,5 @@
 ﻿//----------------------------------------------------------------------
-// AdalJS v1.0.9
+// AdalJS v1.0.9.1
 // @preserve Copyright (c) Microsoft Open Technologies, Inc.
 // All Rights Reserved
 // Apache License 2.0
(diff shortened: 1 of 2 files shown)


tushargupta51:
@brentschmaltz @weijjia @polita

tushargupta51:
Oops, this PR should be against dev. Closing this and opening another one.

Issue #259, opened 2016-04-05
Requirements for iframe redirectUri content


sondreb:
I want to avoid loading the whole web application when the token is renewed in the iframe. There is a lot of JavaScript and processing happening in the main app (it's an SPA), and loading the whole thing in the iframe is something I want to avoid.

So what are the requirements for what to load in the redirectUri?

After user have logged on, I basically do this to change the redirect URL:

`authContext.config.redirectUri = window.location.origin + '/auth.html';`

My problem though is that reading the token from the cache does not work when I load my empty "auth.html" and not "index.html".

`authContext.getCachedToken(resource);`

I have tried to load adal.js, my libs (including jQuery), adal.angular.js, and some of my ADAL initialization code within the auth.html, but it never updates the token in the main frame (app) when loading is completed in the iframe.

So what are the requirements for code that runs in the iframe? Why are there requirements at all really, shouldn't ADAL.js simply read the token from the URL and update in the main app?

Is there any ADAL functions need to call within my auth.html?

Note: It works fine if I don't override the redirectUri, but that means it loads the whole SPA on token renewal.

sondreb:
I duplicated my content from index.html into auth.html, then started removing code that was running. It appears that angular needs to be bootstrapped. Is there any way to avoid this? Right now we have enabled HTML5 mode, which in the current release (1.0.9) means it will do an extra redirect, so on every token renewal my app are loaded twice, and there are megabytes of JavaScript.

PavelPikat:
The problem was raised in #151 and hasn't been fixed yet. As I can see, it was added to 1.0.10 milestone, so all we have to do now is wait :)

sondreb:
Doesn't look like it is the same problem, but might be related. I don't get any errors or problems like that, it simply doesn't update the token cache after loading iframe unless the iframe loads the whole app and runs the Angular initialize for ADAL.

sondreb:
I made a hack that seems to work fine, I would advice against using it, but this is what we are going to do until another solution is available. This is the code I have within my "auth.html", which is a very simple HTML file that only loads adal.js.

The code uses internal methods in ADAL.js, which is probably not smart to do, but had to do it to make it work.

I would like to add a feature request to support a scenario where the iframe loads a separate HTML file and not the whole app. I was expecting ADAL to read the id_token from the iframe when it was done loading, but it doesn't appear to do so? Is that what it's suppose to?

```js
[...]

Issue #151, opened 2015-07-10
Page reloads 3 or  4 times after log-in


drGarbinsky:
There is a strange issue that looks similar to other posted issues but can't quite figure it out. After logging in and being redirected back to the app the page loads 3 or 4 times in a row. After that everything works great. It appears to be triggered by the first call to a protected endpoint via $http.

I have an API on a different URL with CORS configured and oauth2AllowImplicitFlow set to true. 
The general structure of the app is based on the [SinglePageApp-WebAPI-AngularJS-DotNet] (https://github.com/AzureADSamples/SinglePageApp-WebAPI-AngularJS-DotNet) sample.

I posted a issue on [stackoverflow] (http://stackoverflow.com/questions/30768562/app-is-re-initializing-the-first-time-adal-protected-url-is-accessed-via-http) with more detail.

I've been fighting this for a while and would really appreciate any help or troubleshooting guidance.

weijjia:
I'm looking into the issue, and trying to reproduce it on my side.

drGarbinsky:
Appreciate any help. Thanks. Let me know if more details are needed.

weijjia:
Did you take the latest version? I'm playing with the SinglePageApp-WebAPI-AngularJS-DotNet sample, and I could not reproduce the problem by trying: 1) set the require AD to true for the route you want to access, which is triggered by the first call to a protected endpoint via $http 2) click on log in and then do the same thing. 

If you could provide the detailed steps for reproducing, that will be great.

drGarbinsky:
I did take the latest version. Those are the repro steps. I'll go back to a clean copy of the ToDo Spa App sample and see if I can reproduce it as well.

weijjia:
Where is the repro steps? Did you forget to attach?

drGarbinsky:
Sorry The ones that you listed are the basic repro steps: 1) Setup the app. 2) log into the app then do something to trigger a AJAX request to a protected endpoint.

drGarbinsky:
I think I was able to repro the issue using the stock sample. Here are the specific repro steps I used.
1. If logged in, log out of app.
2. Set a breakpoint in the .config function within app.js (this is how I am measuring the behavior)
3. Click "Login" in the upper right
4. Log in
5. Notice that the break point is hit once.
6. Click on ToGo List link (uses a protected endpoint)
7. .config is invoked again.

[...]

PR #284, opened 2016-05-17
Proposed fix for issue #151. New config setting for iframe redirect uri.

Neil665:
This can be used to prevent the situation where the entire application is
re-created inside the iframe.

See further details at:
https://github.com/AzureAD/azure-activedirectory-library-for-js/issues/259#issuecomment-216691463

### DIFF
--- a/lib/adal.js
+++ b/lib/adal.js
@@ -179,7 +179,7 @@ AuthenticationContext.prototype.login = function () {
     this._saveItem(this.CONSTANTS.STORAGE.ERROR_DESCRIPTION, '');
 
 
-    var urlNavigate = this._getNavigateUrl('id_token', null) + '&nonce=' + encodeURIComponent(this._idTokenNonce);
+    var urlNavigate = this._getNavigateUrl('id_token', null, false) + '&nonce=' + encodeURIComponent(this._idTokenNonce);
     this.frameCallInProgress = false;
     this._loginInProgress = true;
     if (this.config.displayCall) {
--- a/lib/adal.js
+++ b/lib/adal.js
@@ -155,6 +155,10 @@ AuthenticationContext = function (config) {
     if (!this.config.redirectUri) {
         this.config.redirectUri = window.location.href;
     }
+
+    if (!this.config.frameRedirectUri) {
+        this.config.frameRedirectUri = this.config.redirectUri;
+    }
 };
 
 /**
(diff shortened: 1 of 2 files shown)


tushargupta51:
@Neil665 Thanks for your contribution :) We would like to merge your changes to our dev branch for next release. It seems you need a rebase.
Also, can you pass the "false" flag in login method where it calls `getNavigateUrl`? I know, its not needed but its good to pass it explicitly.

tushargupta51:
@Neil665 On second look at this PR, this solution will require users to register `frameRedirectUri` as one of the reply address in AAD portal for this to work. In that case, they can just assign the special html  to the `redirectUri` property. In case of login, adal saves the start page and redirect the user back to it once login process is completed. So, the user experience will not change even when the redirect_uri is referring to the special html file.

Neil665:
@tushargupta51 We didn't need to make any change in the AAD portal. Our adal-frame.html file is located under the same path as our main application so it is already covered by the return url that was registered with AAD for the main application.

Neil665:
@tushargupta51 I have made the change that you requested (adding the "false" flag to the method call).
I have not done a rebase as I'm not clear on your exact process, e.g. whether commits should be squashed. Is this documented somewhere?
(Also there was another change we had to make to the library which I have submitted as https://github.com/AzureAD/azure-activedirectory-library-for-js/issues/316).

tushargupta51:
@Neil665 Some internal partners who are using adal js and authenticating against ADFS reported that they needed to add frameRedirect uri as a valid redirect uri for this to work. In any case, why not just assign the special html path to `redirect_uri`?

tushargupta51:
@Neil665  Thanks for you contribution. However, I think `redirectUri` can be used to specify the special frame redirect html and that won't require users to register 2 redirect_uri, both for AAD and ADFS. I have documented the solution here: https://github.com/AzureAD/azure-activedirectory-library-for-js/wiki/FAQs#q1-my-app-is-re-loading-every-time-adal-renews-a-token. Please let me know if that doesn't work.

ryanflowers:
Hello @tushargupta51 I am currently use adal.js "adal-angular": "1.0.17". I am not using the popup option but instead using the full auth redirect on login. I have the same issue listed here in that my app is loading in the iframe. So I created an auth.html which I serve at /auth.html. I overwrite the redirectUri in config post login to be "auth.html" as you described [here](https://github.com/AzureAD/azure-activedirectory-library-for-js/wiki/FAQs#q1-my-app-is-re-loading-every-time-adal-renews-a-token). 

[Issue]
Lets say mid session I get a "login_required" on acquire token due to logout execution within another tab. I handle this by calling login(). The overwritten redirectUri is then used by adal.js on line 1532 which then redirects the user to the /auth.html instead of my app post login.

str.push('redirect_uri=' + encodeURIComponent(obj.redirectUri));

It feels like being explicit on two redirect uri's; one for top and one for iframe is much more straight forward and will lead to less bugs in adding the flexibility needed.

Any input here on direction would be really helpful. Thank you Tushar! BTW I know there is a new MSAL version which I would love to use but I have legacy blockers ATM.

Issue #282, opened 2016-05-10
1.0.10 initialize adal.config.anonymousEndpoints to []


tushargupta51:
Currently, users are forced to add `anonymousEndpoints: [],` to the init configuration. This array should have a default initial value. Add this to adal.js init method

    if (!this.config.anonymousEndpoints) {
        this.config.anonymousEndpoints = [];
    }

dhodgin:
when i use 1.0.10 in a sandbox app i don't have to add anonymousEndpoints: [] to my init but when i dropped adal-angular into my business application i get errors about it being undefined and was forced to add it to the init call. I am using the exact same version in both applications. What's the scoop on this and is this landing in 1.0.11?

tushargupta51:
@dhodgin This will be fixed in 1.0.11 so you don't have to specify default in your app.

Issue #278, opened 2016-05-04
LOGIN_REQUEST URL isn't used when login() is called explicitly 


ischrei:
We are using the Angular service to explicitly invoke login(). We have noticed that the LOGIN_REQUEST value which is saved before redirection to AAD, isn't used in this flow.
It is use in the pure JS version (in handleWindowCallback), and the Angular code only supports the START_PAGE value which is set during automated login which is triggered by a state change.

A suggested fix:
https://github.com/AzureAD/azure-activedirectory-library-for-js/blob/master/lib/adal-angular.js#L134:
```
} else {
	// Support manual login flow where _adal.login() was called explicitly from client code
	var loginRequestUrl = _adal._getItem(_adal.CONSTANTS.STORAGE.LOGIN_REQUEST);
	if (loginRequestUrl) {
		$location.$$parse(loginRequestUrl);
	}
}
```

tushargupta51:
@ischrei Thanks for reporting this. I fixed it by calling `loginHandler()` instead of `_adal.login()` inside the angular login method. This saves the current url in START_PAGE value and rebuild the url when user is redirected back to adal after logging in.

Issue #232, opened 2016-02-16
Wrong redirect url after login in 1.0.8 in adal-angular


kf-radoslawmaziarka:
Hi.

Your last commit broke redirect mechanism after login, when you have nested states.

Let's assume you have such code:
```
$stateProvider
    .state('app', {
        url: '/app',
        abstract: true,
        controller: 'AppController',
        controllerAs:'appVm',
        templateUrl: 'app/app.html',
    });

$stateProvider
	.state('app.address', {
		url: '/address',
		templateUrl: 'app/address/template.html',
		controllerAs: 'atvm',
});
```
Then in _stateChangeHandler_ method in _adal-angular.js_ in 200 line you have:
```
 $location.$$url = toState.url;
```

With nested states property _toState.url_ it's only _'/address'_, not _'/app/address'_ so redirection fails because there is no state with just _'/address'_ path. 

I think that you should get absolute path like that: [http://stackoverflow.com/a/24502548](http://stackoverflow.com/a/24502548)

nigel-dewar:
Hi there, I think I may be experiencing the same issue.

what happens is, when a user logs into my application for the first time, they get redirected to the home route

$stateProvider
// Get Home Page
.state("home", {
url: "/",
controller: "homeController",
controllerAs: "vm",
templateUrl: "app/modules/home/home.html",
requireADLogin: true
})

And there is no problems there at all.

BUT here is a strange behavior

My application sends emails to users, so they can click a url link, and be taken back to a specific location in my app.

something like this .
.state("sessionGroup", {
url: "/workshops/:workshopId/sessionGroups/:sessionGroupId",
controller: "workshopDetailsController",
controllerAs: "vm",
templateUrl: "app/modules/workshops/workshopDetails.html",
requireADLogin: true
})

which provides a URL path like so : http://localhost/#/workshops/2/sessionGroups/96

this works fine if you are logged in already when you click this link and goes straight to the desired URL,

BUT... If you are not logged in to the app and you try to go to this link,

Adal fires up and gets the user to login, and then boom! problem

the url path redirected after the adal login becomes

http://localhost/#/workshops/:workshopId/sessionGroups/:sessionGroupId?workshopId=2&sessionGroupId=96

which is not at all the route I specified. what do I need to do to ensure proper redirection after login?

Is this a bug? And is the same bug that @kf-radoslawmaziarka is experiencing?

rmaziarka:
[...]

Issue #201, opened 2015-12-14
Removal of token hash in html5 mode causes a browser refresh


Hagsten:
When adal-angular.js tries to remove the token hash (#id_token...) in html5 mode the browser (Chrome) performs a refresh. This is not (i guess) the intended behaviour, it should just remove the hash without a refresh. Below is the code that causes this:

```javascript
if ($location.$$html5) {
   $window.location = $window.location.origin + $window.location.pathname;
} else {
    $window.location.hash = '';
}
```

The browser will refresh when location is set without hash, for example setting location to http://mywebsite.com/foo will issue a refresh while http://mywebsite.com/foo# wont.

A suggestion is to use the $location.hash() instead. This will remove the hash when an empty string is sent in, it will keep the hash (# char) however. This should be the intended behaviour.

```javascript
if ($location.$$html5) {
    $location.hash('');
}
```

or perhaps using url and setting origin, path and search but omitting the hash?

```javascript
if ($location.$$html5) {
    $location.url($window.location.origin + $window.location.pathname + $window.location.search);
}
```

mbrand:
I was about to ask the same question, since we're having problems with it, too.
I'm really wondering why you're doing a full page reload with 
```
$window.location = $window.location.origin + $window.location.pathname;
```
when you're in html5-mode, while you won't do a full page reload when your not
```
$window.location.hash = ''; // this doesn't do a reload.
```

katagaStyle:
I was fighting this issue today, too. :(

katagaStyle:
Any news about bug fixing from microsoft? Pretty annoying bug. :(

jonaseriksson84:
I am not in html5 mode, however my application (which resides in a Sharepoint page) issues refreshes because of the `$window.location.hash = '';` line, causing me to lose the Office navbar. 

A workable solution I've found so far is to replace this line with Hagstens `$location.hash('');`, which seems to do the trick with no unnecessary pageloads.

elferone:
This line of code: 
`$window.location = $window.location.origin + $window.location.pathname;`
was causing a refresh in my Angular application, which most of the time screwed up my angular ui router because of the refresh.

Changing to this:
`$location.url($window.location.origin + $window.location.pathname + $window.location.search);`
totally solve my problem.

tushargupta51:
We updated our code to use `$location.hash()` to reset the hash and also use event.PreventDefault to prevent a location change when renewing the token in iframe. That should fix this issue.

PR #300, opened 2016-06-09
Bug fixes for 1.0.11

tushargupta51:
Multiple bug fixes:
1. #282: initialize anonymousEndpoints with a default value
2. #278: replace login() by loginHandler() so that start page is saved
3. #232 and #130: restore the start page url correctly, so that we do not loose any query string parameters
4. #201: update location hash using $location.hash, and use event.preventDefault when renewing token so browser refresh does not happen.

### DIFF
--- a/lib/adal-angular.js
+++ b/lib/adal-angular.js
@@ -125,31 +125,7 @@
                                         // redirect to login requested page
                                         var loginStartPage = _adal._getItem(_adal.CONSTANTS.STORAGE.START_PAGE);
                                         if (loginStartPage) {
-                                            // Check to see if any params were stored
-                                            var paramsJSON = _adal._getItem(_adal.CONSTANTS.STORAGE.START_PAGE_PARAMS);
-                                            if (paramsJSON) {
-                                                // If params were stored redirect to the page and then 
-                                                // initialize the params
-                                                var loginStartPageParams = JSON.parse(paramsJSON);
-                                                for (var param in loginStartPageParams) {
-                                                    var searchStr = ":" + param;
-                                                    // replace all optional parameters first
-                                                    if (loginStartPage.indexOf(searchStr) > -1) {
-                                                        loginStartPage = loginStartPage.replace(searchStr, loginStartPageParams[param]);
-                                                        delete loginStartPageParams[param];     // delete the replaced parameters, otherwise .search() method will include these as query parameters too
-                                                    }
-                                                    else {
-                                                        var querySearchStr1 = "?" + param;
-                                                        var querySearchStr2 = "&" + param;
-                                                        if (loginStartPage.indexOf(querySearchStr1) == -1 && loginStartPage.indexOf(querySearchStr2) == -1) {
-                                                            delete loginStartPageParams[param];     // delete the non-url parameters, otherwise .search() method will include these as query parameters too
-                                                        }
-                                                    }
-                                                }
-                                                $location.url(loginStartPage).search(loginStartPageParams);
-                                            } else {
-                                                $location.url(loginStartPage);
-                                            }
+                                            $location.url(loginStartPage);
                                         }
                                     }, 1);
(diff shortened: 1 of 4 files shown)


tushargupta51:
@polita @weijjia

lib/adal-angular.js
```diff
@@ -220,16 +212,7 @@
...
                         if (isADLoginRequired(toState, _adal.config)) {
-                            if (!_oauthData.isAuthenticated && !_adal._renewActive) {
-                                // $location.$$url is set as the page we are coming from
-                                // Update it so we can store the actual location we want to
-                                // redirect to upon returning
-                                $location.$$url = toState.url;
-
-                                // Parameters are not stored in the url on stateChange so
-                                // we store them
-                                _adal._saveItem(_adal.CONSTANTS.STORAGE.START_PAGE_PARAMS, JSON.stringify(toParams));
```
polita: Have you removed the constant now that we don't use it?

lib/adal-angular.js
```diff
@@ -125,16 +125,7 @@
...
-                                            if (paramsJSON) {
-                                                // If params were stored redirect to the page and then 
-                                                // initialize the params
-                                                var loginStartPageParams = JSON.parse(paramsJSON);
-                                                $location.url(loginStartPage).search(loginStartPageParams);
-                                            } else {
-                                                $location.url(loginStartPage);
-                                            }
+                                            $location.url(loginStartPage);
                                         }
```
polita: We chatted about this, but we should make sure we're saving off and adding back on any hash string.

tushargupta51: Yes, it preserves the hash component too.

lib/adal-angular.js
```diff
@@ -328,33 +311,40 @@
...
-                        } else {
+                        }
+                        else {
                             // Cancel request if login is starting
                             if (authService.loginInProgress()) {
-                                authService.info('login already started.');
-                                return $q.reject('login in progress, cancelling the request');
-                            } else {
+                                authService.info('login is in progress.');
+                                config.data = 'login in progress, cancelling the request';
```
polita: Nit: we should be consistent in our messages: either capitalize the first word for all of them, or have them all lowercase. Below, "Error" is capitalized.

lib/adal.js
```diff
@@ -662,7 +662,7 @@ var AuthenticationContext = (function () {
 
             if (requestInfo.requestType === this.REQUEST_TYPE.LOGIN) {
                 this._loginInProgress = false;
-                this._saveItem(this.CONSTANTS.STORAGE.LOGIN_ERROR, requestInfo.parameters.errorDescription);
+                this._saveItem(this.CONSTANTS.STORAGE.LOGIN_ERROR, requestInfo.parameters.error_description);
```
polita: Why did we add an underscore between "error" and "description"? It looks like the predominant style is to use camel casing.

tushargupta51: It should match with what AAD returns in the url. AAD uses error_description.

polita: Got it, thanks.

lib/adal-angular.js
```diff
@@ -83,11 +83,8 @@
                         var requestInfo = _adal.getRequestInfo(hash);
                         _adal.saveTokenFromHash(requestInfo);
 
-                        if ($location.$$html5) {
-                            $window.location = $window.location.origin + $window.location.pathname;
-                        } else {
-                            $window.location.hash = '';
-                        }
+                        // resetting hash to remove fragments sent by AAD
+                        $location.hash('');
```
weijjia: i remembered that developer were complaining that the hash change will trigger that location event, will this also cause the problem?

tushargupta51: @weijjia $location.hash('') will not trigger another location change. Updating $window.location was doing that.

weijjia:
:shipit:

PR #301, opened 2016-06-10
Added module.exports to expose Logging and AuthenticationContext

rohitnarula7176:
#295


tushargupta51:
@rohitnarula7176 We should have tests for these. Can you add some tests in adalSpec.js?

tushargupta51:
@rohitnarula7176 :shipit: after squashing the commits.

PR #315, opened 2016-06-21
remove redundant constant in angular to save the login start page.

### LINKED ISSUE #313: _adal.CONSTANTS.STORAGE.START_PAGE is redundant
`adal-angular.js.loginHandler()` method saves the current location into `_adal.CONSTANTS.STORAGE.START_PAGE` constant before initializing the login process. `adal.js.login()` method also saves the current location into `_adal.CONSTANTS.STORAGE.LOGIN_REQUEST`.

There is no need to save the start page location in both adal and angular level. Angular wrapper should use the constant set by underlying adal library and redirect the user back to the saved location. Currently, it uses `$location.url('')` to update the location, but that won't work if we want to change the url part before "#". Since post-login redirect is a full page redirect, using `window.location` instead of `$location.url('')` will not cause any additional reload.

Remove the angular level constant and use `window.location` when redirecting users post login.

tushargupta51:
Fixes #313 

Update: also addresses #321  and #316

### DIFF
--- a/lib/adal-angular.js
+++ b/lib/adal-angular.js
@@ -79,12 +79,8 @@
 
                     if (_adal.isCallback(hash)) {
                         // callback can come from login or iframe request
-
                         var requestInfo = _adal.getRequestInfo(hash);
                         _adal.saveTokenFromHash(requestInfo);
-
-                        // resetting hash to remove fragments sent by AAD
-                        $location.hash('');
 
                         if (requestInfo.requestType !== _adal.REQUEST_TYPE.LOGIN) {
                             _adal.callback = $window.parent.AuthenticationContext().callback;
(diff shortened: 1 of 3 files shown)


tushargupta51:
@weijjia @rohitnarula7176 @polita

lib/adal-angular.js
```diff
@@ -123,9 +123,9 @@
                                         updateDataFromCache(_adal.config.loginResource);
                                         $rootScope.userInfo = _oauthData;
                                         // redirect to login requested page
-                                        var loginStartPage = _adal._getItem(_adal.CONSTANTS.STORAGE.START_PAGE);
+                                        var loginStartPage = _adal._getItem(_adal.CONSTANTS.STORAGE.LOGIN_REQUEST);
                                         if (loginStartPage) {
-                                            $location.url(loginStartPage);
+                                            $window.location = loginStartPage;
```
weijjia: so we basically want to trigger the locationchange event for the normal full login redirect?

weijjia:
:shipit:

polita:
:shipit:

lib/adal-angular.js
tushargupta51: The reason I have to use $window.location is because now loginStartPage is the entire url and not just the part after "#" as it was earlier. $location does not allow to reset the part before "#".

PR #317, opened 2016-06-22
add timeout for token renewal requests

### LINKED ISSUE #250: All AcquireToken calls silently fail after ERR_NETWORK_SUSPENDED
If the renew token response called in response to the user's cached access token being expired gets refused due to a **ERR_NETWORK_SUSPENDED** network error, then the acquireToken call silently fails and the callback we passed in remains registered to the auth context. 

In addition, any subsequent calls to acquireToken for the same resource also fail silently because the original callback is still registered so the authentication context instance decides to wait for the original call to complete (which will never happen at this point). 

Hit this issue in both 1.0.7 and 1.0.8.

tushargupta51:
Fixes #250

### DIFF
--- a/tests/unit/spec/AdalSpec.js
+++ b/tests/unit/spec/AdalSpec.js
@@ -155,7 +155,6 @@ describe('Adal', function () {
         adal.config.clientId = 'client';
         adal.config.redirectUri = 'contoso_site';
         spyOn(adal, 'promptUser');
-        console.log('instance:' + adal.instance);
         adal.login();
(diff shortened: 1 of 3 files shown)


tushargupta51:
@dmxfee, this is the PR inspired from your changes in #220

tests/unit/spec/AdalSpec.js
```diff
@@ -269,10 +269,11 @@ describe('Adal', function () {
         expect(adal._renewStates.length).toBe(1);
         // Wait for initial timeout load
         console.log('Waiting for initial timeout');
-        waits(2000);
+        waitsFor(function() {
+ 			return frameMock.src !== 'about:blank' ;
```
polita: Nit: is this indented correctly?

weijjia:
@tushargupta51 As we talked, see if we can capture those network errors and returns it back. Besides this, the change itself is good to me.

weijjia:
:shipit:

PR #320, opened 2016-06-23
update comments in code. fixes #309

### LINKED ISSUE #309: clean up comments in adal-angular.js
There are a few typos in the comments, some variables that are referred to that are not matched in casing, and an invalid comment that no longer applies that make reading the code a bit confusing.

submitting a PR in a few minutes

dhodgin:
There are a few typos in the comments, some variables that are referred to
that are not matched in casing, and an invalid comment that no longer
applies that make reading the code a bit confusing.

The comment about $injector was removed because it appears it no longer
applies. $injector is not used anymore in the service. There is no
circular dependency issue.

### DIFF
--- a/lib/adal-angular.js
+++ b/lib/adal-angular.js
@@ -66,7 +66,7 @@
                     throw new Error('You must set configOptions, when calling init');
                 }
 
-                // loginresource is used to set authenticated status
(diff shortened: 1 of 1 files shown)


tushargupta51:
Thanks @dhodgin , will merge the PR.

PR #322, opened 2016-06-24
Added unit tests for adaljs

rohitnarula7176:
1. Added Logging as a static property of AuthenticationContext.
2. Updated handleWindowCallback function so that the window.Location is not set to the LOGIN page in the case of RENEW_TOKEN request.
3. Added unit tests for handleWindowCallback and renewToken.


tushargupta51:
:shipit:

PR #327, opened 2016-06-27
fixed 312 to generate unique correlationId/request to AAD

rohitnarula7176:


lib/adal.js
```diff
@@ -950,9 +946,10 @@ var AuthenticationContext = (function () {
                 str.push(obj.extraQueryParameter);
             }
 
-            if (obj.correlationId) {
-                str.push('client-request-id=' + encodeURIComponent(obj.correlationId));
-            }
+            if (!obj.correlationId)
+                obj.correlationId = this._guid();
+
```
tushargupta51: This will not generate a new guid for each request. It will generate it the first time and then re-use it.

tushargupta51: Also, please add unit tests for this.

tests/unit/spec/AdalSpec.js
```diff
@@ -276,12 +276,12 @@ describe('Adal', function () {
         console.log('Waiting for initial timeout');
         waitsFor(function () {
             return mockFrames['adalRenewFrame' + RESOURCE1].src !== 'about:blank';
-        }, 'iframe src not updated', 2000);
+    }, 'iframe src not updated', 2000);
```
tushargupta51: indentation

lib/adal.js
```diff
@@ -1151,7 +1146,7 @@ var AuthenticationContext = (function () {
 
     AuthenticationContext.prototype.log = function (level, message, error) {
         if (level <= AuthenticationContext.Logging.level) {
-            var correlationId = this.config.correlationId;
+            var correlationId = this.config.correlationId ? this.config.correlationId : this._guid();
```
tushargupta51: as we discussed, add correlation id only if the user set it, otherwise not.

PR #330, opened 2016-06-28
Set active renewals to null before invoke callback

hejiantao5:


tushargupta51:
@hejiantao5 What is the reason for this change? Are you running into an issue?

hejiantao5:
@tushargupta51 yes, the issue is the _activeRenewals won't be reset to null and all acquireToken will run into registerCallback and window.callBacksMappedToRenewStates will become bigger and bigger but window.callBackMappedToRenewStates will never invoke

tushargupta51:
@hejiantao5 I agree that if `window.callBackMappedToRenewStates` is not invoked, the callbacks remain registered and subsequent calls to acquireToken will just register yet another callback. But that is the intended behavior. There is an open issue https://github.com/AzureAD/azure-activedirectory-library-for-js/issues/250 related to this, where these callbacks are left registered if the original request is never completed (because of network error or something). We are fixing that in 1.0.11 by adding timeout to token renewal request.

In your change, the _activeRenewals will not be set to null if window.callBackMappedToRenewStates is not invoked. The only change you have made is move up the setting of activeRenewal property which will not make any difference if window.callBackMappedToRenewStates is called (as that property will be set to null once all callbacks are executed).

hejiantao5:
@tushargupta51 I saw #250, but what I encounter situation is window.callBackMappedToRenewStates is called but some errors happened in my callback function, then all other acquireToken won't work

tushargupta51:
@hejiantao5 If one of the callbacks throws, these 2 statements will not be executed:

             window.callBacksMappedToRenewStates[expectedState] = null;
              window.callBackMappedToRenewStates[expectedState] = null;

When acquireToken method is called next time, new callbacks will get registered to the existing list, and the callback which threw an error earlier will be called before the new callback. I think the better fix is to have a try..catch pattern.

tushargupta51:
@hejiantao5 We need to use try...catch pattern here instead of moving the lines above the callbacks calling code. I have opened an issue to track this: #399 . I am closing this PR as there are merge conflicts and this PR is against master. If you are interested in fixing this, please rebranch from dev and submit the PR.

PR #337, opened 2016-07-02
[adal]: add ability to switch off navigation to the LOGIN_REQUEST url

spzSource:
After calling `handleWindowCallback()` user is moved to `LOGIN_REQUEST` url. And it is not dependent of what redirect url we specify for application through azure AD portal.

Since redirect url may differ from `LOGIN_REQUEST`, it will be good to make such behaviour customisable.

I will be happy for feedback and merge of suggested pull-request.


spzSource:
@tushargupta51, @brandwe can you, guys, make review?

tushargupta51:
@spzSource Why do you want to redirect the user back to redirectUri instead of the page from where they started the login process? I am trying to understand the scenario where redirecting back to login_start page is not the right/expected behavior.

spzSource:
@tushargupta51, thanks for feedback.
Let me describe my use-cases:

1) Lets imagine that I have a requirement according to I have the separate page for login. Only after successful  authentication I have ability to move to the main page.

So, my configuration looks like this:
```
Login page: https://localhost:3300/login/
Redirect handler: https://localhost:3300/login-handler/
Main page: https://localhost:3300/main

```
And after handleWindowCallback() execution on the "Redirect handler" page, user navigates to login page again instead of navigating to the main page.

2) The second use case relates on acquiring token for more than one app.

My configuration looks like following:
```
endpoints: {
        'http://localhost:11605/api': '81263916-12321-123-9659-fff5ad33f1a4',
        'http://localhost:55252/api': '123123ds-3a5c-sad-234234-f3e7652f6b'
}
```
And when user trying to acquire token for second resource (http://localhost:55252/api, for instance), adal.js creates shadow iframe and retrieves a token by performing automatic login in scope of shadow iframe. In this case it completely breaks down navigation mechanism in angular 2, because in scope of shadow iframe 'router'(angular2 navigation provider) navigates to login page instead of appropriate page that represent the http://localhost:55252/api resource.

lib/adal.js
```diff
@@ -147,6 +147,9 @@ AuthenticationContext = function (config) {
 
     this.config = this._cloneConfig(config);
 
+    // by default we need to move to LOGIN_REQUEST url after window callback handling.
+    this.config.navigateToLoginRequestUrl = true;
```
rohitnarula7176: Since navigateToLoginRequestUrl is defined on the config object, setting it to true like this in the constructor, will overwrite the value passed by the user in the config object when the constructor is called.
Could you please change it to something like:
if (typeof this.config.navigateToLoginRequestUrl === 'undefined')
            this.config.navigateToLoginRequestUrl = true;
By doing this, if the value is undefined, we set it to true or else we use whatever value is passed to us by the user.

lib/adal.js
```diff
@@ -788,7 +791,10 @@ AuthenticationContext.prototype.handleWindowCallback = function () {
         }
 
         window.location.hash = '';
-        window.location = this._getItem(this.CONSTANTS.STORAGE.LOGIN_REQUEST);
+        if (this.config.navigateToLoginRequestUrl) {
```
rohitnarula7176: If the user sets navigateToLoginRequestUrl to false, then after the redirect, the token will be left in the hash part of the fragment which is not safe. We would just like some additional checks here:
1. Do not set window.location if we are already on the same page to prevent an app reload. 
2. In other cases, just set the hash part of the url to null.
Please refer to the snippet below for your reference:
  if (window.location.href.indexOf('#') != -1)
                    var urlWithoutHash = window.location.href.split("#")[0];
                else
                    var urlWithoutHash = window.location.href;
                if (this.config.navigateToLoginRequestUrl && urlWithoutHash !== this._getItem(this.CONSTANTS.STORAGE.LOGIN_REQUEST))
                    window.location = this._getItem(this.CONSTANTS.STORAGE.LOGIN_REQUEST);
                else
                    window.location.hash = '';

rohitnarula7176:
@spzSource Thank you for the pull request. I reviewed your PR and have made some comments. Please review them when you get a chance. You will have to rebase your branch with dev since we just made another release last week. If you are comfortable making changes to your PR, please let us know or else we will make the necessary changes.

spzSource:
Hi, @rohitnarula7176 thanks for comments. I will try to make changes mentioned above.

spzSource:
Sorry, I didn't notice that  changes have already been made.

tushargupta51:
@spzSource I think there are some issues in your rebase. The number of changes seems huge. Also, this needs to go against dev branch instead of master. @rohitnarula7176 is working on similar change updating adal-angular js too: #390 . That PR actually took your changes and added some more stuff on top of it. If you are fine, we can close this PR, and review the other one and get it merged to dev.

spzSource:
@tushargupta51 Yes, I have already noticed that @rohitnarula7176 already working on this stuff. So ,yes it's fine. You can close this PR. I think, you guys will do appropriate changes more efficiently than me. :)

Cheers, A

tushargupta51:
Thanks @spzSource  for your contribution!

PR #343, opened 2016-07-13
fixed regex for extracting url fragment params

rohitnarula7176:


tests/unit/spec/AdalSpec.js
```diff
@@ -807,6 +811,41 @@ describe('Adal', function () {
...
+        expect(obj.state).toBe(STATE);
+        expect(obj.session_state).toBe(SESSION_STATE);
+
+        obj = adal._deserialize(INVALID_URLFRAGMENT);
+        expect(obj.id_token).toBeUndefined;
+        expect(obj.state).toBe(STATE);
+        expect(obj.session_state).toBe(SESSION_STATE);
+        expect(obj['id_token' + IDTOKEN_MOCK]).toBeUndefined;
+
+        adal._deserialize = function (query) {
```
tushargupta51: There is an issue with this pattern. Any tests that we add below this test will use the incorrect "deserialize" method as it is being overridden here. You can save the old function and reassign it to adal._deserialize.

PR #347, opened 2016-07-19
fixed authorization header overrides so we can use different kinds of…

brechtbilliet:
This fix makes sure that it will not set the authorization header when  it is already set. Some services (mostly api's) require other authorization headers like "basic"

### DIFF
--- a/lib/adal-angular.js
+++ b/lib/adal-angular.js
@@ -338,7 +338,9 @@ if (typeof module !== 'undefined' && module.exports) {
                         if (tokenStored) {
                             authService.info('Token is avaliable for this url ' + config.url);
                             // check endpoint mapping if provided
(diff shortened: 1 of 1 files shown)


lib/adal-angular.js
```diff
@@ -338,7 +338,9 @@ if (typeof module !== 'undefined' && module.exports) {
                         if (tokenStored) {
                             authService.info('Token is avaliable for this url ' + config.url);
                             // check endpoint mapping if provided
-                            config.headers.Authorization = 'Bearer ' + tokenStored;
+                            if(config.headers.Authorization){
```
tushargupta51: I think you want the reverse check i.e. when `config.headers.Authorization` is not set, then set it, otherwise not.

Issue #60, opened 2014-12-20
Add support for authentication using pop-up method instead of full redirect. 


mattmazzola:
**I think adaljs should add support for authentication using a pop-up instead of a full redirect.**

There are two major benefits to using this technique:
* Much easier for developers!
* Works with SPA frameworks.

# Much easier for developers!
With the pop-up method more of the conceptual complexity can be abstracted into the adaljs library and the developer can have much easier time using the library.  Adaljs currently does a full redirect and then in order to work with angularjs router there is a bunch of code like:
```
$rootScope.$on('$locationChangeStart', locationChangeHandler);
....
var locationChangeHandler = function () {
                var hash = $window.location.hash;

                if (_adal.isCallback(hash)) {
                    // callback can come from login or iframe request

                    var requestInfo = _adal.getRequestInfo(hash);
                    _adal.saveTokenFromHash(requestInfo);
....
```

This is watching the url and if it has a parameter of 'id_token' then it saves the tokens etc.. This logic is currently exposed to the framework such as AngularJS and would have to re-written for every framework which is the largest barrier to using adaljs. Now imagine you have to rewrite all of this for Ember, React, Knockout, etc... way too much overhead especially when people have to try to integrate into existing projects which may already have complex router configurations.

I hope in the future adaljs is modeled after other major JS OAuth SDK's such as Facebook or Google.
FB: http://connect.facebook.net/en_US/all.js
Google: https://apis.google.com/js/client.js?onload=googleApiLoaded

With pop-up style of authentication you can get rid of all url watching stuff in angular and have a simple call like: 
```
FB.login(function(fbResponse) {
  if (fbResponse.authResponse) {
    Ember.run(function() {
      resolve({ accessToken: fbResponse.authResponse.accessToken });
    });
  } else {
    reject();
  }
});
```
This encapsulates the entire login process in one asynchronous call! Yes, with pop-ups there is a the possibility that pop-ups maybe disabled, but the benefits out weigh this greatly. Especially if this is an option the developer can choose.  Just realize what that means for ease of use! FB/Google SDK's handle setting up the window and watching that frame's url for change and closing then it has redirected.  I assume this is extremely similar to how the Windows 8 WebAuthenticationBroker works.

# Works with SPA frameworks.
[...]

PR #357, opened 2016-07-27
Rohitnarula7176/popup

rohitnarula7176:
#60

### DIFF
--- a/lib/adal-angular.js
+++ b/lib/adal-angular.js
@@ -84,34 +84,28 @@
                         var requestInfo = _adal.getRequestInfo(hash);
                         _adal.saveTokenFromHash(requestInfo);
 
-                        if (requestInfo.requestType !== _adal.REQUEST_TYPE.LOGIN) {
-                            _adal.callback = $window.parent.AuthenticationContext().callback;
-                            if (requestInfo.requestType === _adal.REQUEST_TYPE.RENEW_TOKEN) {
-                                _adal.callback = $window.parent.callBackMappedToRenewStates[requestInfo.stateResponse];
-                            }
-                        }
-
                         // Return to callback if it is sent from iframe
                         if (requestInfo.stateMatch) {
-                            if (typeof _adal.callback === 'function') {
+                            if (requestInfo.requestType === _adal.REQUEST_TYPE.RENEW_TOKEN) {
+                               var callback = $window.parent.callBackMappedToRenewStates[requestInfo.stateResponse];
                                 // since this is a token renewal request in iFrame, we don't need to proceed with the location change.
                                 event.preventDefault();
 
                                 // Call within the same context without full page redirect keeps the callback
-                                if (requestInfo.requestType === _adal.REQUEST_TYPE.RENEW_TOKEN) {
+                                if (callback && typeof callback === 'function') {
                                     // id_token or access_token can be renewed
                                     if (requestInfo.parameters['access_token']) {
-                                        _adal.callback(_adal._getItem(_adal.CONSTANTS.STORAGE.ERROR_DESCRIPTION), requestInfo.parameters['access_token']);
+                                        callback(_adal._getItem(_adal.CONSTANTS.STORAGE.ERROR_DESCRIPTION), requestInfo.parameters['access_token']);
                                         return;
                                     } else if (requestInfo.parameters['id_token']) {
-                                        _adal.callback(_adal._getItem(_adal.CONSTANTS.STORAGE.ERROR_DESCRIPTION), requestInfo.parameters['id_token']);
+                                        callback(_adal._getItem(_adal.CONSTANTS.STORAGE.ERROR_DESCRIPTION), requestInfo.parameters['id_token']);
                                         return;
                                     } else if (requestInfo.parameters['error']) {
-                                        _adal.callback(_adal._getItem(_adal.CONSTANTS.STORAGE.ERROR_DESCRIPTION), null);
+                                        callback(_adal._getItem(_adal.CONSTANTS.STORAGE.ERROR_DESCRIPTION), null);
                                         return;
                                     }
                                 }
-                            } else {
+                            } else if(requestInfo.requestType === _adal.REQUEST_TYPE.LOGIN){
                                 // normal full login redirect happened on the page
                                 updateDataFromCache(_adal.config.loginResource);
                                 if (_oauthData.userName) {
@@ -121,10 +115,13 @@
                                         $rootScope.userInfo = _oauthData;
                                     }, 1);
 
-                                    $rootScope.$broadcast('adal:loginSuccess');
+                                    $rootScope.$broadcast('adal:loginSuccess', _adal._getItem(_adal.CONSTANTS.STORAGE.IDTOKEN));
                                 } else {
                                     $rootScope.$broadcast('adal:loginFailure', _adal._getItem(_adal.CONSTANTS.STORAGE.ERROR_DESCRIPTION));
                                 }
+
+                                if (_adal.callback && typeof _adal.callback === 'function')
+                                    _adal.callback(_adal._getItem(_adal.CONSTANTS.STORAGE.ERROR_DESCRIPTION), _adal._getItem(_adal.CONSTANTS.STORAGE.IDTOKEN));
 
                                 // redirect to login start page
                                 var loginStartPage = _adal._getItem(_adal.CONSTANTS.STORAGE.LOGIN_REQUEST);
--- a/lib/adal-angular.js
+++ b/lib/adal-angular.js
@@ -87,7 +87,7 @@
                         // Return to callback if it is sent from iframe
                         if (requestInfo.stateMatch) {
                             if (requestInfo.requestType === _adal.REQUEST_TYPE.RENEW_TOKEN) {
-                               var callback = $window.parent.callBackMappedToRenewStates[requestInfo.stateResponse];
+                                var callback = $window.parent.callBackMappedToRenewStates[requestInfo.stateResponse];
                                 // since this is a token renewal request in iFrame, we don't need to proceed with the location change.
                                 event.preventDefault();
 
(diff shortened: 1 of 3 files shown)


lib/adal.js
```diff
@@ -189,6 +200,28 @@ var AuthenticationContext = (function () {
         // callback from redirected page will receive fragment. It needs to call oauth2Callback
     };
 
+    AuthenticationContext.prototype.loginPopup = function (urlNavigate) {
+        var popupWindow = window.open(urlNavigate, "login", 'width=483, height=600');
```
polita: Should we try to position this somewhere other than upper left? It would be nice to center/middle it.

lib/adal.js
```diff
@@ -189,6 +200,28 @@ var AuthenticationContext = (function () {
...
+        if (popupWindow && popupWindow.focus)
+            popupWindow.focus();
+        var that = this;
+        var registeredRedirectUri = this.config.redirectUri;
+        var pollTimer = window.setInterval(function () {
+            if (!popupWindow || popupWindow.closed || popupWindow.closed === undefined) {
+                window.clearInterval(pollTimer);
+            }
+            try {
+                if (popupWindow.document.URL.indexOf(registeredRedirectUri) != -1) {
```
polita: Why check for the redirectUri in the popup window URL? Is there an attack we're preventing?

tests/unit/spec/AdalSpec.js
```diff
@@ -182,7 +182,7 @@ describe('Adal', function () {
 
     it('calls displaycall if given for login', function () {
         storageFake.setItem(adal.CONSTANTS.STORAGE.USERNAME, 'test user');
-
+        adal._loginInProgress = false;
```
polita: Why do the tests set the private var here?

lib/adal.js
```diff
@@ -165,6 +170,8 @@ var AuthenticationContext = (function () {
      */
     AuthenticationContext.prototype.login = function () {
         // Token is not present and user needs to login
+        if (this._loginInProgress)
+            return;
```
tushargupta51: log an info level message here.

lib/adal.js
```diff
@@ -189,6 +200,28 @@ var AuthenticationContext = (function () {
...
     };
 
+    AuthenticationContext.prototype.loginPopup = function (urlNavigate) {
+        var popupWindow = window.open(urlNavigate, "login", 'width=483, height=600');
+        if (popupWindow && popupWindow.focus)
+            popupWindow.focus();
+        var that = this;
+        var registeredRedirectUri = this.config.redirectUri;
+        var pollTimer = window.setInterval(function () {
+            if (!popupWindow || popupWindow.closed || popupWindow.closed === undefined) {
```
tushargupta51: Why do we need the last check: popupWindow.closed === undefined?

lib/adal.js
```diff
@@ -189,6 +200,28 @@ var AuthenticationContext = (function () {
...
+                window.clearInterval(pollTimer);
+            }
+            try {
+                if (popupWindow.document.URL.indexOf(registeredRedirectUri) != -1) {
+                    window.clearInterval(pollTimer);
+                    window.location.hash = popupWindow.location.hash;
```
tushargupta51: We should not reset the window.location.hash as this will cause app reload. Instead, update handleWindowCallback method to take hash as parameter.

lib/adal.js
```diff
@@ -189,6 +200,28 @@ var AuthenticationContext = (function () {
...
+                    that._loginInProgress = false;
```
tushargupta51: No need to reset this flag here. This gets reset in saveTokenFromHash.

lib/adal.js
```diff
@@ -810,6 +843,9 @@ var AuthenticationContext = (function () {
                 this.verbose('Window is redirecting');
                 callback = this.callback;
             }
+            callback = this.callback;
```
tushargupta51: Why set this callback here? I think we need to understand the if checks: window && window.oauth2Callback. That else if block should execute for login requests.

lib/adal.js
```diff
@@ -810,6 +843,9 @@ var AuthenticationContext = (function () {
...
+            if (callback)
+                callback(this._getItem(this.CONSTANTS.STORAGE.ERROR_DESCRIPTION), requestInfo.parameters[this.CONSTANTS.ACCESS_TOKEN] || requestInfo.parameters[this.CONSTANTS.ID_TOKEN]);
             window.location = this._getItem(this.CONSTANTS.STORAGE.LOGIN_REQUEST);
```
tushargupta51: if popUp is set to true, we should not redirect the main app to login start page, as it is already on that page.

lib/adal.js
```diff
@@ -810,6 +843,9 @@ var AuthenticationContext = (function () {
                 this.verbose('Window is redirecting');
                 callback = this.callback;
             }
+            callback = this.callback;
+            if (callback)
+                callback(this._getItem(this.CONSTANTS.STORAGE.ERROR_DESCRIPTION), requestInfo.parameters[this.CONSTANTS.ACCESS_TOKEN] || requestInfo.parameters[this.CONSTANTS.ID_TOKEN]);
```
tushargupta51: Related to above comment. This callback is for login request and I think we should move it to the else if block. Also, we should only pass id_token to the callback in this case, no access_token.

lib/adal.js
polita: Please do provide a top and left. On a multimon setup, this popped up the window on the other monitor for me.

lib/adal.js
```diff
@@ -189,6 +200,48 @@ var AuthenticationContext = (function () {
...
+        return popupWindow;
+    }
+
+    AuthenticationContext.prototype.loginPopup = function (urlNavigate) {
+        var popupWindow = this.openPopup(urlNavigate, "login", 483, 600);
+        if (this.config.redirectUri.indexOf('#') != -1)
+            var registeredRedirectUri = this.config.redirectUri.split("#")[0];
+        else
+            var registeredRedirectUri = this.config.redirectUri;
+        var that = this;
```
tushargupta51: This is not needed now.

lib/adal.js
```diff
@@ -790,10 +843,11 @@ var AuthenticationContext = (function () {
     };
 
     /*exported  oauth2Callback */
-    AuthenticationContext.prototype.handleWindowCallback = function () {
+    AuthenticationContext.prototype.handleWindowCallback = function (hash) {
```
tushargupta51: This change is no longer needed now.

lib/adal.js
```diff
@@ -189,6 +200,48 @@ var AuthenticationContext = (function () {
         // callback from redirected page will receive fragment. It needs to call oauth2Callback
     };
 
+    AuthenticationContext.prototype.openPopup = function (urlNavigate, title, w, h) {
```
tushargupta51: Lets define this as a private method (even though it will still be on prototype). Also, please add comments to explain what this code is doing as it is assuming some variables to be defined in the current context.

tushargupta51:
As we discussed, please add:
1. Logging
2. condition for not setting window.location in adal-angular when config.popUp is true
3. Add tests

lib/adal-angular.js
```diff
@@ -344,9 +342,22 @@
...
                             if (authService.loginInProgress()) {
-                                authService.info('login is in progress.');
-                                config.data = 'login in progress, cancelling the request for ' + config.url;
-                                return $q.reject(config);
+                                if (authService.config.popUp) {
+                                    authService.info('Url: ' + config.url + ' will be loaded after login is successful');
+                                    var delayedRequest = $q.defer();
+                                    $rootScope.$on('adal:loginSuccess', function (event, token) {
+                                        if (token) {
+                                            config.headers.Authorization = 'Bearer ' + tokenStored;
```
tushargupta51: Add a log statement here too: "Login Completed. Sending request for " + config.url.

lib/adal.js
```diff
@@ -175,18 +185,62 @@ var AuthenticationContext = (function () {
...
+            var top = ((height / 2) - (popUpHeight / 2)) + winTop;
+
+            var popupWindow = window.open(urlNavigate, title, 'width=' + popUpWidth + ', height=' + popUpHeight + ', top=' + top + ', left=' + left);
+            if (popupWindow.focus) {
+                popupWindow.focus();
+            }
+            return popupWindow;
+        } catch (e) {
+            this.info(e.message);
+            this.loginInProgress = false;
```
tushargupta51: this should be this._loginProgress.

lib/adal.js
```diff
@@ -175,18 +185,62 @@ var AuthenticationContext = (function () {
...
+            return null;
+        }
+    }
+
+    AuthenticationContext.prototype._loginPopup = function (urlNavigate) {
+        var popupWindow = this._openPopup(urlNavigate, "login", 483, 600);
```
tushargupta51: since we return null from _openPopup and also, in some browsers window.open won't work, we should check if 'popupWindow' is null, if so, log an error message and set the LOGIN_ERROR constant.

lib/adal.js
```diff
@@ -175,18 +185,62 @@ var AuthenticationContext = (function () {
...
+        var that = this;
+        var pollTimer = window.setInterval(function () {
+            if (!popupWindow || popupWindow.closed || popupWindow.closed === undefined) {
+                that._loginInProgress = false;
+                window.clearInterval(pollTimer);
+            }
+            try {
+                if (popupWindow.document.URL.indexOf(that.config.redirectUri) != -1) {
```
tushargupta51: Why removed the check for '#'? This will cause the pop up to not close when using plain JS and the redirect url has '#' in it.

lib/adal.js
```diff
@@ -175,18 +185,62 @@ var AuthenticationContext = (function () {
...
+                    window.location.hash = popupWindow.location.hash;
+                    window.clearInterval(pollTimer);
+                    popupWindow.close();
+                }
+            } catch (e) {
```
tushargupta51: log this error using that.error.

lib/adal-angular.js
```diff
@@ -135,7 +132,8 @@
                                     if (!$location.$$html5 && loginStartPage.indexOf('#') > -1) {
```
tushargupta51: Please move the log statement and this if statement inside the if block for adal.popUp

lib/adal-angular.js
tushargupta51: @rohitnarula7176 I think you missed this one.

lib/adal.js
tushargupta51: We should define the constants "483" and "600" at the top.

PR #358, opened 2016-08-04
First implementation of angular2 http and service

damienpontifex:
A http subclass for adal to intercept requests and add tokens plus an angular2 service that wraps the adal.js base similar to the existing setup for adal-angular for angular1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@
 .externalToolBuilders
 *.swp
 node_modules
+typings
 *~
 /.c9revisions/
 doc/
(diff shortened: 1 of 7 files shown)


tushargupta51:
@damienpontifex Thanks a lot for you contribution! But unfortunately, we are not making any feature changes to adal now. We are focusing our efforts on next generation of this library - MSAL, that will work against AAD converged endpoint (v2.0). MSAL will be written in TypeScript and we are considering angular 2 for that.

Issue #360, opened 2016-08-05
adal:loginFailure should not mask error messages from AAD


hawkerm:
It seems adal consumes some of the messages returned by the AAD service, so it can be hard to know what state the system is in if there's an error authenticating.

For instance in adal-angular.js it broadcasts the loginFailure with 'auto renew failure' from the locationChangeHandler eating the original error returned.

Issue #175, opened 2015-10-11
Token Signature Validation


Bartmax:
Is the token signature validated ?

weijjia:
Nope, adal.js currently doesn't do token signature validation, but this is something we're planning to add in the library.

Bartmax:
Thanks. I'm leaving this open for you to tracking. Feel free to close it.

brentschmaltz:
adal does not validate tokens.

PR #362, opened 2016-08-10
Add token validation callback

tushargupta51:
Partially addresses the #175 by providing an extensibility point for users to plug in for client side token validation.

Also, fixing #360 by passing the error to application when auto renewal of id token fails


tushargupta51:
Not adding this feature now.

Issue #345, opened 2016-07-14
anonymousEndpoints feature with interceptor enabled fails with loading 2 templateUrls at once


max-buecker:
![deadlock](https://cloud.githubusercontent.com/assets/19929742/16840498/5242109a-49d5-11e6-8108-45c9c864080f.png)
Running adal and adal-angular with ui-router introduces a problem.
When states have templateUrls defined, and you jump from level 1 state, into a nested level 3 state, two templates need to be loaded first. In my case `/host-application/module-container.html` and `/modules/vacation/views/list.html`

While some stuff in adal-angular pushes the `/modules/vacation/views/list.html` url into the anonymousEndpoints array, adal.js at about line 748 checks for the resource of `/host-application/module-container.html`, which hasn't been added to anonymousEndpoints yet, marking it as `config.loginResource`.

The stateChange is not triggered, because it waits for the template to be resolved, therefor the login doesn't trigger, which is supposed to continue the delayed call for the template. So it's kind of a deadlock situation:

The router waits for the template, the template waits for the login, the login waits for the stateChange in the router.

Heres a screenshot. I do not have time to make a minimal reproduction project.
The scenario in the screenshot is navigating from the state `landingpage`(templateUrl: "/host-application/landingpage.html") to the state `modules.vacation.list`(templateUrl: "/modules/vacation/views/list.html")

The extra layer from state `modules.vacation` doesn't require a call, as its template is defined statically:
```javascript
        $stateProvider.state("modules.vacation", {
                url: "/modules/vacation",
                template: "<div class='module-content slideLeft-animation' ui-view></div>",
                requireADLogin: true
            }).state("modules.vacation.list", {
                url: "/modules/vacation/list",
                templateUrl: "/modules/vacation/views/list.html",
                controller: "vacation.listController",
                requireADLogin: true
            }).state("modules.vacation.create", {
                url: "/modules/vacation/create",
                templateUrl: "/modules/vacation/views/create.html",
                controller: "vacation.createController",
                requireADLogin: true
            });
        $stateProvider.state("landingpage", {
            url: "/landingpage",
            templateUrl: "/host-application/landingpage.html",
            controller: "host-application.landingpage"
        }).state("modules", {
            url: "/modules",
            templateUrl: "/host-application/module-container.html",
            controller: "host-application.module-container"
        });
```

tushargupta51:
[...]

PR #363, opened 2016-08-11
code to handle anonymous endpoints and nested states

rohitnarula7176:
#345

### DIFF
--- a/lib/adal-angular.js
+++ b/lib/adal-angular.js
@@ -208,10 +208,16 @@
                             }
                         }
                         else {
-                            if (typeof nextRoute.$$route.templateUrl === 'function')
-                                nextRoute.$$route.templateUrl = nextRoute.$$route.templateUrl(nextRoute.params);
-                            if (nextRoute.$$route.templateUrl && !isAnonymousEndpoint(nextRoute.$$route.templateUrl)) {
-                                _adal.config.anonymousEndpoints.push(nextRoute.$$route.templateUrl);
+                            if (nextRoute.$$route.templateUrl) {
+                                var urlNavigate = '';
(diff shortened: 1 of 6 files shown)


lib/adal-angular.js
```diff
@@ -198,6 +199,16 @@
                 }
 
                 var routeChangeHandler = function (e, nextRoute) {
+                    if (!_paths) {
+                        _paths = $injector.get('$route').routes;
+                        for (var route in _paths) {
+                            if (_paths[route].hasOwnProperty('templateUrl') && !isADLoginRequired(_paths[route], _adal.config)) {
+                                if (!(_adal.config.anonymousEndpoints.indexOf(_paths[route].templateUrl) > -1)) {
```
tushargupta51: use isAnonymousEndpoint(url) method here. Its possible that user added "App/Views" to anonymous endpoints in which case we don't need to add "App/Views/Home.html" as every route under "App/Views" is considered anonymous. Your check does a full string match.

lib/adal-angular.js
```diff
@@ -207,15 +218,19 @@
...
-                        }
                     }
                 };
 
                 var stateChangeHandler = function (e, toState, toParams, fromState, fromParams) {
+                    if (!_paths) {
+                        _paths = $injector.get('$state').get();
+                        _paths.forEach(function (state) {
+                            if (state.templateUrl && !isADLoginRequired(state, _adal.config)) {
+                                if (!(_adal.config.anonymousEndpoints.indexOf(state.templateUrl) > -1))
```
tushargupta51: same comment as above. Use isAnonymousEndpoint method.

rohitnarula7176:
#368

Issue #88, opened 2015-02-24
Question about Nonce


manfredsteyer:
Hi,

I’ve noticed, that you create the nonce with JavaScript using Math.rand(). Therefore, the nonce is not random in a cryptographic way. However, as Crypto-API is quire young and not widely supported, there is no other way to do this in JavaScript. One alternative would be to contact a Web API for creating a nonce in a cryptographic random way. 

Now I’m curious, whether you considered this solution and why you choose to use Math.rand().

Wishes,
Manfred

omercs:
Thank you for feedback. Adal is generating GUID version4 and it is not crypto random. We are looking into that.

tushargupta51:
We now use window.cyrpto api to generate guids, if it is supported by browser. If not, we fall back to Math.random. Fixed in 1.0.12.

PR #367, opened 2016-08-19
added window.crypto to generate _guid

rohitnarula7176:
#88

### DIFF
--- a/lib/adal.js
+++ b/lib/adal.js
@@ -1082,29 +1082,43 @@ var AuthenticationContext = (function () {
         // Format: xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx
         // y could be 1000, 1001, 1010, 1011 since most significant two bits needs to be 10
         // y values are 8, 9, A, B
-        var guidHolder = 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx';
-        var hex = '0123456789abcdef';
-        var r = 0;
-        var guidResponse = "";
-        for (var i = 0; i < 36; i++) {
-            if (guidHolder[i] !== '-' && guidHolder[i] !== '4') {
-                // each x and y needs to be random
-                r = Math.random() * 16 | 0;
+        var cryptoObj = window.crypto || window.msCrypto; // for IE 11
+        if (cryptoObj && cryptoObj.getRandomValues) {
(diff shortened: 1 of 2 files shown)


lib/adal.js
```diff
@@ -1082,29 +1082,43 @@ var AuthenticationContext = (function () {
...
-                guidResponse += hex[r];
-            } else if (guidHolder[i] === 'y') {
-                // clock-seq-and-reserved first hex is filtered and remaining hex values are random
-                r &= 0x3; // bit and with 0011 to set pos 2 to zero ?0??
-                r |= 0x8; // set pos 3 to 1 as 1???
-                guidResponse += hex[r];
-            } else {
-                guidResponse += guidHolder[i];
+            return decimalToHex(buffer[0]) + decimalToHex(buffer[1]) + '-' + decimalToHex(buffer[2]) + '-' + decimalToHex(buffer[3]) + '-' +
+              decimalToHex(buffer[4]) + '-' + decimalToHex(buffer[5]) + decimalToHex(buffer[6]) + decimalToHex(buffer[7]);
```
tushargupta51: This is not in the format: xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx. Can you check on the significance of '4' and 'y'. It seems to be related to rfc 4122

lib/adal.js
```diff
@@ -1084,17 +1084,21 @@ var AuthenticationContext = (function () {
         // y values are 8, 9, A, B
         var cryptoObj = window.crypto || window.msCrypto; // for IE 11
         if (cryptoObj && cryptoObj.getRandomValues) {
-            var buffer = new Uint16Array(8);
+            var buffer = new Uint8Array(16);
             cryptoObj.getRandomValues(buffer);
+            buffer[6] |= 0x40;
```
tushargupta51: This looks good to me, please add some comments here explaining how you are setting bits, this will help someone reading the code.

e.g. buffer[6] and buffer[7] represents time_hi_and_version field. We will update the most significant 4 bits of buffer[6] to represent decimal number 4 (uuid version number).
buffer[6] | 01000000 will set the 7 bit to 1
buffer[6] & 01001111 will set the 5, 6, and 8 bit to 0 such that bit 5-8 == 0100 = "4"

tushargupta51:
:shipit:

PR #369, opened 2016-08-22
add support for templateUrl that is a function. fixes #368

### LINKED ISSUE #368: bug in isAnonymousEndpoint() if templateUrl is a function instead of string
per the angular documentation the templateUrl can be either a string or a function. if it is a function it is passed the params of the route

see: https://docs.angularjs.org/api/ngRoute/provider/$routeProvider

in the adal-angular code
```
function isAnonymousEndpoint(url)
```

there is a url.indexOf() call. If your templateUrl for this route is a function you will get a console error that url.indexOf is not a function because url is a function that will return a string if involked with the params of the upcoming route.

we use a function for one of our routes to manipulate one of the params so the page title can be dynamically updated using the param value.

everytime we hit this route we get this error.

I'm attempting to patch this by checking if the url is a function and then involking it first to get the string value back before calling the .indexOf() but i'm having trouble getting access to the params of the upcoming route to involk the templateUrl function.

i've tried $routeParams and $route.current.params as per angular docs to try to acces the params for the route but i'm either getting nothing or the current route

the other option is to change the `routeChangeHandler` and `stateChangeHandler` functions and check the params there before calling the `isAnonymousEndpoint` function all together.

as far as i can see those are the only 2 places `isAnonymousEndpoint` is called in adal-angular so it could be easy to modify

i've also tried adding the path to the template to my anonymous endpoints array in my init but it doesn't stop the error because it's still going to end up passing the function reference to `isAnonymousEndpoint` in the handler code regardless.

dhodgin:
The templateUrl property for the routeProvider object can be a string or a function as per angular docs.

if it is a string it returns a path to a html template. if it is a function it is passed the params for the route and returns a string to a template.

adal-angular does not currently handle the case where the templateUrl is a function

NOTE: this patch only fixes the problem for the routeChangeHandler. The stateChangeHandler will also have this problem but only for people using ui-router. I have never used ui-router and don't know how the stateChangeHandler works or is called. i have no way in my project to trigger this event handler to test if toParams is the same type of object or not. Someone will have to write a patch in another PR to handle when the toState.templateUrl is a function. otherwise they will get url.indexOf is not a function errors on state change.

### DIFF
--- a/lib/adal-angular.js
+++ b/lib/adal-angular.js
@@ -198,6 +198,8 @@
                 }
 
                 var routeChangeHandler = function (e, nextRoute) {
+                    var nextRouteUrl;
+
                     if (nextRoute && nextRoute.$$route) {
                         if (isADLoginRequired(nextRoute.$$route, _adal.config)) {
                             if (!_oauthData.isAuthenticated) {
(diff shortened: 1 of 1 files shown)


tushargupta51:
@dhodgin  Thanks for your contribution. @rohitnarula7176 is going to add support for ui-router.

dhodgin:
Thanks. sounds good.

PR #370, opened 2016-08-26
added anonymous endpoints feature

rohitnarula7176:
#345

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -13,10 +13,10 @@ module.exports = function(config) {
       'bower_components/angular-mocks/angular-mocks.js',
       'bower_components/angular-route/angular-route.js',
       'bower_components/angular-resource/angular-resource.js',
-	  'bower_components/angular-ui-router/release/angular-ui-router.js',
+      'bower_components/angular-ui-router/release/angular-ui-router.js',
       'lib/*.js',
       'tests/testApp.js',
-	  'tests/stateApp.js',
+      'tests/stateApp.js',
       'tests/angularModuleSpec.js'
     ],
 
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -13,8 +13,10 @@ module.exports = function(config) {
       'bower_components/angular-mocks/angular-mocks.js',
       'bower_components/angular-route/angular-route.js',
       'bower_components/angular-resource/angular-resource.js',
+	  'bower_components/angular-ui-router/release/angular-ui-router.js',
       'lib/*.js',
       'tests/testApp.js',
+	  'tests/stateApp.js',
       'tests/angularModuleSpec.js'
     ],
 
--- a/lib/adal-angular.js
+++ b/lib/adal-angular.js
@@ -226,8 +226,6 @@
                 }
 
                 var routeChangeHandler = function (e, nextRoute) {
-                    var nextRouteUrl;
-
                     if (nextRoute && nextRoute.$$route) {
                         if (isADLoginRequired(nextRoute.$$route, _adal.config)) {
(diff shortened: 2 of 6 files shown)


karma.conf.js
```diff
@@ -13,8 +13,10 @@ module.exports = function(config) {
       'bower_components/angular-mocks/angular-mocks.js',
       'bower_components/angular-route/angular-route.js',
       'bower_components/angular-resource/angular-resource.js',
+	  'bower_components/angular-ui-router/release/angular-ui-router.js',
```
tushargupta51: indentation

karma.conf.js
```diff
@@ -13,8 +13,10 @@ module.exports = function(config) {
...
       'lib/*.js',
       'tests/testApp.js',
+	  'tests/stateApp.js',
```
tushargupta51: indentation

lib/adal-angular.js
```diff
@@ -197,9 +197,35 @@
...
-                var routeChangeHandler = function (e, nextRoute) {
-                    var nextRouteUrl;
+                function getStates(toState) {
+                    var state = null;
+                    var states = [];
+                    if (toState.hasOwnProperty('parent')) {
+                        state = toState;
+                        while (state) {
+                            states.unshift(state);
+                            state = $injector.get('$state').get(state.parent);
```
tushargupta51: what happens when state.parent is null?

rohitnarula7176: when state.parent is null, state = $injector.get('$state').get(state.parent) makes state= null and we exit the while loop.

lib/adal-angular.js
```diff
@@ -197,9 +197,35 @@
...
+                        if (state) {
+                            states.push(state);
+                        }
+                        for (var i = 1; i < stateNames.length; i++) {
+                            stateName += '.' + stateNames[i];
+                            state = $injector.get('$state').get(stateName);
+                            if (state) {
+                                states.push(state);
+                            }
+                        }
```
tushargupta51: You can do something like this here to reduce code redundancy:

                       for (var i = 0, stateName = stateNames[0]; i < stateNames.length; i++) {
                            state = $injector.get('$state').get(stateName);
                            if (state) {
                                states.push(state);
                            }
                            stateName += '.' + stateNames[i];
                        }

tests/angularModuleSpec.js
```diff
@@ -397,3 +397,60 @@ describe('TaskCtl', function () {
...
+        $templateCache.put('profile.html', '');
+        $templateCache.put('settings.html', '');
+        $templateCache.put('account.html', '');
+        $templateCache.put('name.html', '');
+        adalServiceProvider.config.anonymousEndpoints = [];
+    }));
+
+    it('checks if anonymous endpoints are populated on statechange event if states are nested and separated by .', function () {
+        rootScope.$on('$stateChangeSuccess', function (event, toState) {
+            state = toState;
```
tushargupta51: `var state` ?

rohitnarula7176: state is defined in the before each block.

karma.conf.js
tushargupta51: @rohitnarula7176 you missed this...if you carefully look, you will notice its a tab in the front of the line instead of a space

tests/angularModuleSpec.js
tushargupta51: That is a reference to $state, we should not update that. use a local variable to assign "toState" to.

tushargupta51:
:shipit:

PR #390, opened 2016-09-16
switch off navigation to Login_Request Url

rohitnarula7176:
#337

### DIFF
--- a/lib/adal-angular.js
+++ b/lib/adal-angular.js
@@ -69,14 +69,26 @@
 
                 var locationChangeHandler = function (event, newUrl, oldUrl) {
                     _adal.verbose('Location change event from ' + oldUrl + ' to ' + newUrl);
-                    var hash = $window.location.hash;
+                    if ($location.$$html5) {
+                        var hash = $location.hash();
+                    }
+                    else {
+                        var hash = '#' + $location.path();
+                    }
+                    processHash(hash);
 
+                    $timeout(function () {
+                        updateDataFromCache(_adal.config.loginResource);
+                        $rootScope.userInfo = _oauthData;
+                    }, 1);
+                };
+
+                var processHash = function (hash) {
                     if (_adal.isCallback(hash)) {
                         // callback can come from login or iframe request
                         _adal.verbose('Processing the hash: ' + hash);
                         var requestInfo = _adal.getRequestInfo(hash);
                         _adal.saveTokenFromHash(requestInfo);
-
                         // Return to callback if it is sent from iframe
                         if (requestInfo.stateMatch) {
                             if (requestInfo.requestType === _adal.REQUEST_TYPE.RENEW_TOKEN) {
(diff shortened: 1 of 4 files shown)


lib/adal-angular.js
```diff
@@ -127,15 +127,19 @@
...
+                                            _adal.verbose('Redirecting to start page: ' + loginStartPage);
+                                            if (!$location.$$html5 && loginStartPage.indexOf('#') > -1) {
+                                                $location.url(loginStartPage.substring(loginStartPage.indexOf('#') + 1));
+                                            }
+                                            $window.location = loginStartPage;
                                         }
-                                        $window.location = loginStartPage;
                                     }
+                                    else
+                                        $window.location.hash = '';
```
tushargupta51: I remember some users reported that this caused a reload for them. Can you please what happens when we reset only the hash?

lib/adal-angular.js
```diff
@@ -127,15 +127,19 @@
...
-                                            $location.url(loginStartPage.substring(loginStartPage.indexOf('#') + 1));
+                                    if (_adal.config.navigateToLoginRequestUrl) {
+                                        var loginStartPage = _adal._getItem(_adal.CONSTANTS.STORAGE.LOGIN_REQUEST);
+                                        if (loginStartPage) {
+                                            // prevent the current location change and redirect the user back to the login start page
+                                            _adal.verbose('Redirecting to start page: ' + loginStartPage);
+                                            if (!$location.$$html5 && loginStartPage.indexOf('#') > -1) {
+                                                $location.url(loginStartPage.substring(loginStartPage.indexOf('#') + 1));
+                                            }
+                                            $window.location = loginStartPage;
```
tushargupta51: This behavior is little different from plain adal, where we do not reload if loginStartPage is same as window.location (after removing the hash). Here we always reload.

lib/adal-angular.js
```diff
@@ -127,15 +127,19 @@
...
                                 if (!_adal.popUp) {
-                                    var loginStartPage = _adal._getItem(_adal.CONSTANTS.STORAGE.LOGIN_REQUEST);
-                                    if (loginStartPage) {
-                                        // prevent the current location change and redirect the user back to the login start page
-                                        _adal.verbose('Redirecting to start page: ' + loginStartPage);
-                                        if (!$location.$$html5 && loginStartPage.indexOf('#') > -1) {
-                                            $location.url(loginStartPage.substring(loginStartPage.indexOf('#') + 1));
+                                    if (_adal.config.navigateToLoginRequestUrl) {
+                                        var loginStartPage = _adal._getItem(_adal.CONSTANTS.STORAGE.LOGIN_REQUEST);
+                                        if (loginStartPage) {
```
tushargupta51: If loginStartPage is "", we will not make any changes to the url. I think if loginStartPage is "", we should atleast reset the hash. Also, see if we can reduce some nesting here.

tushargupta51:
@rohitnarula7176 general comment: please look at these issues once: https://github.com/AzureAD/azure-activedirectory-library-for-js/issues?q=is%3Aissue+location.hash+is%3Aclosed. People ran into some issues with resetting window.location.hash. Just make sure, we are not re-introducing old bugs.

lib/adal.js
```diff
@@ -278,10 +281,11 @@ var AuthenticationContext = (function () {
         }, 20);
     };
 
-    /**
-     * Checks if login is in progress.
-     * @returns {Boolean} true if login is in progress, false otherwise.
-     */
+    AuthenticationContext.prototype._onPopUpHashChanged = function (hash) {
+        var evt = new CustomEvent('popUpHashChanged', { detail: hash });
```
tushargupta51: Please update the event name to 'adal:popUpHashChanged'.

lib/adal-angular.js
tushargupta51: please change this to window.location.href

PR #394, opened 2016-09-27
Stringify the object when logging

rohitpagariya:
In adal-angular.js, there is an instance where error is concatenated with a string causing the object to be printed as "[Object object]". This fix stringifies the error object before logging.

I also added a stringified details to the log when an error response is received in adal-angular


tushargupta51:
@rohitpagariya Thanks for the contribution! The changes look good to me. I will merge the PR.

rohitpagariya:
Thanks.

PR #395, opened 2016-09-27
fix: unreachable return statement

rohitpagariya:
The return statement at the end of the function is unreachable due to
the if-else control flow


tushargupta51:
@rohitpagariya Thanks Rohit for your contribution! This looks good to me.

PR #403, opened 2016-10-03
use adfs specific logout url when tenant is adfs

### LINKED ISSUE #389: logOut does not work on ADFS
When calling `logOut` against on-premise adfs browser is redirected to `https://<host>/adfs/oauth2/logout?post_logout_redirect_uri=<redirect_uri>` which is not supported by adfs (yet?).

The correct logout uri for adfs would be `https://<host>/adfs/ls/?wa=wsignout1.0`

What do you think about adding conditional logout uri depending on `tenant` being equal to `adfs` or not?

NickAb:
Fixes #389 

Overrides logout navigation url if tenant is adfs.

### DIFF
--- a/lib/adal.js
+++ b/lib/adal.js
@@ -546,11 +546,11 @@ var AuthenticationContext = (function () {
         }
 
         var urlNavigate = this.instance + tenant + '/oauth2/logout?' + logout;
-		
-		if (tenant === 'adfs') {
-			urlNavigate = this.instance + '/adfs/ls/?wa=wsignout1.0'
-		}
-		
+
+        if (this.config.logOutUri) {
+            urlNavigate = this.instance + this.config.logOutUri;
+        }
+
         this.info('Logout navigate to: ' + urlNavigate);
         this.promptUser(urlNavigate);
     };
--- a/lib/adal.js
+++ b/lib/adal.js
@@ -534,21 +534,24 @@ var AuthenticationContext = (function () {
      */
     AuthenticationContext.prototype.logOut = function () {
         this.clearCache();
-        var tenant = 'common';
-        var logout = '';
         this._user = null;
-        if (this.config.tenant) {
-            tenant = this.config.tenant;
-        }
-
-        if (this.config.postLogoutRedirectUri) {
-            logout = 'post_logout_redirect_uri=' + encodeURIComponent(this.config.postLogoutRedirectUri);
-        }
-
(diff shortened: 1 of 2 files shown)


NickAb:
I am not sure about using tenant as single condition for adfs logout url as this behavior might be fixed in later adfs releases, which will make it version dependent.
Maybe a new config property `logoutUrl` should be introduced, which, when set, will override default navigation url. Any thoughts?

tushargupta51:
@NickAb Hi Nick, thanks for your contribution. I think a new config property to allow users to set the logOut url seems better approach as we cannot rely on "adfs" as valid tenant value (its not really a tenant).

tushargupta51:
@NickAb We are getting closer to our next release date and would like to include this fix as well. Would you be able to update PR sometime this week or early next week? If you cannot take time out for this, no worries, let us know, we will make changes on top of your commits, if possible.

NickAb:
@tushargupta51 I've made required changes. I added `logOutUri` option to config. If set it will be used together with `instance` option to build logout url. Note that it does not use `tenant`, I think it will be more flexible that way, as user would be able to include tenant by self into the `logOutUri` option if needed.

lib/adal.js
```diff
@@ -546,6 +546,11 @@ var AuthenticationContext = (function () {
         }
 
         var urlNavigate = this.instance + tenant + '/oauth2/logout?' + logout;
+
+        if (this.config.logOutUri) {
+            urlNavigate = this.instance + this.config.logOutUri;
+        }
+
         this.info('Logout navigate to: ' + urlNavigate);
```
tushargupta51: @NickAb I think logOutUri should be an absolute uri like redirectUri. Users configuring this already knows about the instance. We should use it as is. Also, can we structure this a little differently like this:

        var urlNavigate; 
        if (this.config.logOutUri) {
             urlNavigate = this.config.logOutUri;
         }
         else {
              if (this.config.postLogoutRedirectUri) {
                  logout = 'post_logout_redirect_uri=' + encodeURIComponent(this.config.postLogoutRedirectUri);
              }
             urlNavigate = this.instance + tenant + '/oauth2/logout?' + logout;
         }

NickAb:
@tushargupta51 fixed according to suggestion. Let me know if any changes needed or if I should squash it commits, etc.

tushargupta51:
@NickAb Thanks for your contribution! The changes look good to me, merged them to our dev branch.

Issue #366, opened 2016-08-16
App reloading inside iFrame when renewing token - Solution 2


Tsunade90:
I'm working on the project for which that issue was posted:

https://github.com/AzureAD/azure-activedirectory-library-for-js/issues/298#issuecomment-225035980

We upgrated our adal.js version to 1.0.11 and use the workaround described here:
https://gist.github.com/tushargupta51/5fa6d000357120adbc8fd1d5c68853c4

It solves our issues in localhost but not when the code is minified/uglyfied and deployed on a distant server, the page stays stuck on frameRedirect.html when the token is trying to be renewed:
http://mysite.azurewebsites.net/frameRedirect.html#/id_token=eyJ0eXAiOiJKV1Q[...]

When debbuging the _locationChangeHandler_ (adal-angular.js file) function, it appears that the variable _loginStartPage_ is always empty (_var loginStartPage = _adal._getItem(_adal.CONSTANTS.STORAGE.LOGIN_REQUEST);_) so the user can't be redirected to the app's home page after the token has been renewed. 
If I set manually a value inside _loginStartPage_ to see what happens, the app reloads forever.

We downgraded the adal version to 1.0.7 for the moment because our app is in production environment so the user can't face a white empty page but we can't keep this version as it causes weird behaviours (the js files are loaded 5-6 times).

Thank you

tushargupta51:
@Tsunade90 Is this issue specific to some browser? Or does it occur on all browsers: Chrome, Firefox, IE/Edge?

Tsunade90:
It occurs on all browsers, sorry for not mentioning it

Tsunade90:
If it helps: seems to have something to do with the "angular components" architecture because if I don't use a redirectUri in my adal config, I get the error:

Error: [$compile:tpload] Failed to load template: scripts/components/master/master-page.html (HTTP status: undefined undefined)
http://errors.angularjs.org/1.5.7/$compile/tpload?p0=scripts%2Fcomponents%2Fmaster%2Fmaster-page.html&p1=undefined&p2=undefined

index.html:
<body>
   <my-main-component></my-main-component>

tushargupta51:
@Tsunade90 Sorry for the delay in getting back on this. Can you please share the special frame html that you have, adal init configuration and route configuration? That will help me debug this further. Also, console logs will help too. You can turn on logging by adding this globally:

Logging = {
    level: 3,
    log: function (message) {
        console.log(message);
    }
[...]

Issue #373, opened 2016-08-29
Login is not working when localLoginUrl is specified


mohanskumar:
adalAuthenticationService.login() function is not working when localLoginUrl is specified.  This is happening in 1.0.11 version.

tushargupta51:
@mohanskumar What is the error or behavior when you call login() method without specifying localLoginUrl? What does your adal configuration code look like?

mohanskumar:
No errors thought. ( nothing happens when the login method is called(); ), wondering the else part is not working in the adal-angular. 
 var loginHandler = function () {
                    _adal.info('Login event for:' + $location.$$url);
                    if (_adal.config && _adal.config.localLoginUrl) {
                        $location.path(_adal.config.localLoginUrl);
                    }
                    else {
                        // directly start login flow
                        _adal.info('Start login at:' + window.location.href);
                        $rootScope.$broadcast('adal:loginRedirect');
                        _adal.login();
                    }
                };

I am currently running v1.0.8 and trying to upgrade to the latest. 1.0.8 works fine so far.

tushargupta51:
@mohanskumar In the config above, you do have a localLoginUrl. Did you comment it out and it didn't work? Can you try debugging to see if its even going to adal.login() method. Because, if its going to the adal.login method, it should redirect to AAD login page.

mohanskumar:
Yes, It works when localLogin URL commented.

tushargupta51:
@mohanskumar I am confused whether it works with localLoginUrl or without localLoginUrl.
Earlier you said this:
> adalAuthenticationService.login() function is not working when localLoginUrl is not specified. This is happening in 1.0.11 version.

In your comment above, you said this:
> Yes, It works when localLogin URL commented.

mohanskumar:
Oops sorry that was a typo. It doesn't work when login url is specified.

tushargupta51:
@mohanskumar What does your routing configuration looks like? Does the "/login" route has "requireADLogin" set to true?

The if block in the code:

    if (_adal.config && _adal.config.localLoginUrl) {
       $location.path(_adal.config.localLoginUrl);
    }

This will redirect the user to "/login" route. Do you see that happening? Please check the network trace in Chrome Developer tools to confirm that.

mohanskumar:
Login route is not set with RequireADLogin to true. As i mentioned with my
issue log, My project works perfectly fine with 1.0.8. and I don't have any
other config/ route settings changed.

[...]

PR #408, opened 2016-10-04
Rohitnarula7176/issue 373

rohitnarula7176:
#373 #360

### DIFF
--- a/tests/angularModuleSpec.js
+++ b/tests/angularModuleSpec.js
@@ -76,6 +76,8 @@ describe('TaskCtl', function () {
             return q.when(token);
         };
 
+        window.parent.AuthenticationContext = window.AuthenticationContext;
+
         controller('TaskCtl', { $scope: scope, adalAuthenticationService: adalServiceProvider });
     }));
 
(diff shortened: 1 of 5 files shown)


tushargupta51:
Also, fixes #366

tushargupta51:
@rohitnarula7176 The changes look good to me. Just one thing: I think we do not have a test where we redirect the user to localLoginUrl. It might be little tricky to write that. You will have to redirect the user to a route that requires authentication or found some way to call into `loginHandler`.

tests/angularModuleSpec.js
```diff
@@ -210,12 +210,9 @@ describe('TaskCtl', function () {
     });
 
     it('tests stateMismatch broadcast when state does not match', function () {
-        window.parent.AuthenticationContext = function () {
-            return {
-                callback: function () { },
-                _renewStates: {}
-            };
-        };
+        window.parent.AuthenticationContext = window.AuthenticationContext;
```
tushargupta51: This creates dependency between 2 tests. So either we should have this line in all the tests below or move it to beforeEach. I am fine with both.

tushargupta51:
:shipit:

PR #410, opened 2016-10-05
Add Support for Additional OpenID Connect Identity Providers

### LINKED ISSUE #409: Enhance to Support Additional Identity Providers
I would like to see this library enhanced to support additional identity providers.  With some fairly minor adjustments this should be possible.

Specifically, I would like to see the following:

- Provide for more flexibility in the instance URL by allowing for a blank tenant and an alternate (or blank) context root (instead of the hard-coded oauth2 value).
- Allow optional scope and response type configuration parameters.
- Allow for identity providers that pass back the id_token as a query string parameter instead of a URL fragment.  This is fairly common with id_tokens; as opposed to responses that include access_tokens.

davewoodward:
Enhanced library to allow for other OpenIDConnect identity providers by supporting blank tenant and custom root context values in the instance URL and optional scope and response type parameters.  Also allowed for the ID_TOKEN to be returned as a URL query parameter rather than a URL fragment.  Finally, added the necessary tests to verify the various scenarios around this functionality.

Closes #409

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,5 +1,5 @@
 {
-  "bitwise": true,
+  "bitwise": false,
   "camelcase": true,
   "curly": true,
   "eqeqeq": true,
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -22,19 +22,14 @@ module.exports = function (grunt) {
         },
         jasmine_node: {
             options: {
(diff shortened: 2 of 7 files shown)


tushargupta51:
Closing this based on our discussion here: #409 .

Issue #372, opened 2016-08-26
Is there a scope event for token renewal?


MaJaHa95:
I have a web call that I want to make every time a user becomes authenticated (essentially a "get notifications" type of thing).

My initial code looked something like this:

```
if ($rootScope.userInfo.isAuthenticated) {
    setup();
}
else {
    $rootScope.$on("adal:loginSuccess", () => {
        setup();
    });
}
```

Basically, if adal has magically found authentication information in `sessionStorage`, then we're good, otherwise, wait for the user to be authenticated.

This works well for most cases, but never renewals. From what I can tell, `adal:loginSuccess` doesn't occur after a renewal is successfully completed, so neither of these trigger.

I've resolved this with a total stop-gap of setting up a `$watch` on `userInfo.isAuthenticated`, but there must be a better way!

Does ADAL trigger an event for renewals, and if not, would that be a logical addition? I don't imagine it should be the same as initial login, but `adal:tokenRenewed` (or something to that effect) would be helpful.

tushargupta51:
@MaJaHa95 Currently, adal does not broadcast events for token renewal operations. I want to understand your scenario a little better. I understand that you want to execute some actions when a user is authenticated, but why do you need to do the same when a token is renewed as that is not a new authentication, its renewing token for a user who was authenticated earlier.

MaJaHa95:
I store some state in memory (it's Angular, if that wasn't assumed).

Basically,

1. User visits website
2. Adal checks for tokens, none found.
3. Initial setup continues, and `userInfo.isAuthenticated` is `false`, so we hold off.
4. Adal logs in, and calls `adal:loginSuccess`, thus queuing `getNotifications()`
 1. We store those results in `$scope`.
5. ...
6. User closes laptop and goes home (or whatever)
7. User comes back, website is revitalized, memory cleared.
8. Adal finds an expired token
9. Initial setup continues, `userInfo.isAuthenticated` is `false`, so we hold off again.
10. Adal runs through its iframe flow to renew a token
11. *Without a notification, I don't know that it's now time to call `getNotifications()` again.*

I could be misunderstanding some aspect of the flow of login--don't hesitate to correct me.

I will admit, I'm not sure of a time when `adal:loginSuccess` is necessary, but another for token renewal wouldn't be. So perhaps I am misunderstanding stuff.


Thanks!
Matthew

rohitpagariya:
[...]

PR #414, opened 2016-10-11
adding scope event for token renewal

rohitnarula7176:
#372

### DIFF
--- a/tests/angularModuleSpec.js
+++ b/tests/angularModuleSpec.js
@@ -422,3 +422,64 @@ describe('StateCtrl', function () {
         }
     });
 });
+
+describe('AcquireTokenCtl', function () {
+    var scope,adalServiceProvider, rootScope, controller,window;
+    var store = {};
+    //mock Application to allow us to inject our own dependencies
+    beforeEach(angular.mock.module('TestApplication'));
+
+    //mock the controller for the same reason and include $scope and $controller
+    beforeEach(angular.mock.inject(function (_adalAuthenticationService_, _$rootScope_, _$controller_, _$window_) {
+        adalServiceProvider = _adalAuthenticationService_;
+        rootScope = _$rootScope_;
+        controller = _$controller_;
+        window = _$window_;
+        //create an empty scope
+        scope = rootScope.$new();
+
+        spyOn(sessionStorage, 'getItem').andCallFake(function (key) {
+            return store[key];
+        });
+        spyOn(sessionStorage, 'setItem').andCallFake(function (key, value) {
+            store[key] = value;
+        });
+        spyOn(window, 'Date').andCallFake(function () {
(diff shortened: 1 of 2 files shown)


tests/angularModuleSpec.js
```diff
@@ -422,3 +422,60 @@ describe('StateCtrl', function () {
...
 });
+
+describe('TaskCtl', function () {
+    var scope,adalServiceProvider, rootScope, controller,window;
+    var store = {};
+    //mock Application to allow us to inject our own dependencies
+    beforeEach(angular.mock.module('TestApplication'));
+
+    //mock the controller for the same reason and include $scope and $controller
+    beforeEach(angular.mock.inject(function (_adalAuthenticationService_, _$rootScope_, _$controller_, _$httpBackend_, _$q_, _$window_, _$route_, _$location_) {
```
tushargupta51: You can take off some of the values from the signature that are not used.

tests/angularModuleSpec.js
```diff
@@ -422,3 +422,60 @@ describe('StateCtrl', function () {
...
+    afterEach(function () {
+        store = {};
+    });
+
+    it('checks if acquireTokenSuccess/acquireTokenFailure events are broadcasted in case of acquireToken', function () {
+        var resource = adalServiceProvider.config.loginResource;
+        var token = 'token123';
+        spyOn(rootScope, '$broadcast').andCallThrough();
+        scope.$on('adal:acquireTokenFailure', function (event, message) {
+            expect(message).toBe('User login is required');
```
tushargupta51: We do not check if the events are called or not. This test will pass even when adal:acquireTokenFailure and adal:acquireTokenSuccess events are not called. We should either assign value of "message" to a variable defined in outside scope and test the value of that variable outside of the event listener. Or we should use "haveBeenCalled()" api to ensure that the event was called.

tests/angularModuleSpec.js
```diff
@@ -461,21 +461,27 @@ describe('TaskCtl', function () {
...
+        expect(error).toBe('User login is required');
         store = {
             'adal.token.keys': resource + '|',
-            'adal.access.token.keyloginResource123': resource + 'token123',
+            'adal.access.token.keyloginResource123': token,
             'adal.expiration.keyloginResource123': 122
         };
         scope.$on('adal:acquireTokenSuccess', function (event, message) {
-            expect(message).toBe(resource + token);
+            expect(message).toBe(token);
```
tushargupta51: you dont need expect statements here.

tests/angularModuleSpec.js
tushargupta51: What does it need the $q, $route and $location for?

tushargupta51:
:shipit:

PR #416, opened 2016-10-14
TS Version of adal

con-ssc:
Added yarn lock to ignore
added typescript transcription for adal

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,3 +94,4 @@ samples/owin/OwinSample/Scripts/adal.js
 
 #codecoverage
 coverage
+/yarn.lock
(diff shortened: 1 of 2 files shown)


tisawyer:
Is this PR going to get merged? Having a TypeScript version of Adal would be very helpful for our project.

tushargupta51:
@con-ssc Some community members added ts definition for adal-angular here: https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/adal. Do we need complete TS translation of adal? Is there a case where the typed definition will not work?

con-ssc:
First, it would be easier to keep the definitions up to date with a ts version because you would just emit the typings with it, second, you could make an importable npm module or import it into typescript as you like. Third, Microsoft has very high knowledge in the area of TS like projects like VSCode and PowerBI Visuals and so on. Four, you can emit to a target EcmaScript Version that your environment needs. I could go on for a while :)

Issue #412, opened 2016-10-06
Add API documentation


tushargupta51:
Some of our public apis do not have proper/correct comments. Make sure all apis have comments in the right format that explain the parameters, return values and the purpose of property/method.

The second step will be to generate documentation using "jsdoc". We already have grunt task for that. We should be able to generate documentation using that and upload on github wiki.

tushargupta51:
For reference: http://stackoverflow.com/questions/39261085/azure-ad-adaljs-how-to-use-this-library

brentschmaltz:
@tushargupta51 if this PR was NOT committed, please get the PR in.

tushargupta51:
Let's keep it open until we have this released in 1.0.14.

PR #422, opened 2016-10-20
adding api doc for adaljs

rohitnarula7176:
#412

### DIFF
--- a/lib/adal.js
+++ b/lib/adal.js
@@ -22,29 +22,29 @@ var AuthenticationContext = (function () {
     'use strict';
 
     /**
-     * Config information
-     * @public
-     * @class Config
-     * @property {tenant}          Your target tenant
-     * @property {clientId}        Identifier assigned to your app by Azure Active Directory
-     * @property {redirectUri}     Endpoint at which you expect to receive tokens
-     * @property {instance}        Azure Active Directory Instance(default:https://login.microsoftonline.com/)
-     * @property {endpoints}       Collection of {Endpoint-ResourceId} used for autmatically attaching tokens in webApi calls
-     */
-
-    /**
-     * User information from idtoken.
-     *  @class User
-     *  @property {string} userName - username assigned from upn or email.
-     *  @property {object} profile - properties parsed from idtoken.
+     * Configuration options for Authentication Context.
+     * @class config
+     *  @property {string} tenant - Your target tenant.
+     *  @property {string} clientID - Client ID assigned to your app by Azure Active Directory.
+     *  @property {string} redirectUri - Endpoint at which you expect to receive tokens.Defaults to `window.location.href`.
+     *  @property {string} instance - Azure Active Directory Instance.Defaults to `https://login.microsoftonline.com/`.
+     *  @property {Array} endpoints - Collection of {Endpoint-ResourceId} used for automatically attaching tokens in webApi calls.
+     *  @property {Boolean} popUp - Set this to true to enable login in a popup winodow instead of a full redirect.Defaults to `false`.
+     *  @property {string} localLoginUrl - Set this to redirect the user to a custom login page.
+     *  @property {function} displayCall - User defined function of handling the navigation to Azure AD authorization endpoint in case of login. Defaults to 'null'.
+     *  @property {string} postLogoutRedirectUri - Redirects the user to postLogoutRedirectUri after logout. Defaults to 'null'.
+     *  @property {string} cacheLocation - Sets browser storage to either 'localStorage' or sessionStorage'. Defaults to 'sessionStorage'.
+     *  @property {Array.<string>} anonymousEndpoints Array of keywords or URI's. Adal will not attach a token to outgoing requests that have these keywords or uri. Defaults to 'null'.
+     *  @property {number} expireOffsetSeconds If the cached token is about to be expired in the expireOffsetSeconds (in seconds), Adal will renew the token instead of using the cached token. Defaults to 120 seconds.
+     *  @property {string} correlationId Unique identifier used to map the request with the response. Defaults to RFC4122 version 4 guid (128 bits).
      */
 
     /**
      * Creates a new AuthenticationContext object.
      * @constructor
-     * @param {object}  config               Configuration options for AuthenticationContext
-     *
-     **/
+     * @param {config}  config               Configuration options for AuthenticationContext
+     */
+
     AuthenticationContext = function (config) {
         /**
          * Enum for request type
@@ -169,8 +169,7 @@ var AuthenticationContext = (function () {
     };
 
     /**
-     * Gets initial Idtoken for the app backend
-     * Saves the resulting Idtoken in localStorage.
+     * Initiates the login process by redirecting the user to Azure AD authorization endpoint.
      */
     AuthenticationContext.prototype.login = function () {
         // Token is not present and user needs to login
@@ -202,6 +201,10 @@ var AuthenticationContext = (function () {
         }
     };
 
+    /**
+     * Configures popup window for login.
+     * @ignore
+     */
     AuthenticationContext.prototype._openPopup = function (urlNavigate, title, popUpWidth, popUpHeight) {
         try {
             /**
@@ -231,6 +234,11 @@ var AuthenticationContext = (function () {
         }
     }
 
+    /**
+     * After authorization, the user will be sent to your specified redirect_uri with the user's bearer token
+     * attached to the URI fragment as an id_token field. It closes popup window after redirection.
+     * @ignore
+     */
     AuthenticationContext.prototype._loginPopup = function (urlNavigate) {
         var popupWindow = this._openPopup(urlNavigate, "login", this.CONSTANTS.POPUP_WIDTH, this.CONSTANTS.POPUP_HEIGHT);
         if (popupWindow == null) {
@@ -270,19 +278,28 @@ var AuthenticationContext = (function () {
         }, 20);
     };
 
+    /**
+     * Checks if login is in progress.
+     * @returns {Boolean} true if login is in progress, false otherwise.
+     */
     AuthenticationContext.prototype.loginInProgress = function () {
         return this._loginInProgress;
     };
 
+    /**
+     * Checks for the resource in the cache. By default, cache location is Session Storage
+     * @ignore
+     * @returns {Boolean} 'true' if login is in progress, else returns 'false'.
+     */
     AuthenticationContext.prototype._hasResource = function (key) {
         var keys = this._getItem(this.CONSTANTS.STORAGE.TOKEN_KEYS);
         return keys && !this._isEmpty(keys) && (keys.indexOf(key + this.CONSTANTS.RESOURCE_DELIMETER) > -1);
     };
 
     /**
-     * Gets token for the specified resource from local storage cache
-     * @param {string}   resource A URI that identifies the resource for which the token is valid.
-     * @returns {string} token if exists and not expired or null
+     * Gets token for the specified resource from the cache.
+     * @param {string}   resource A URI that identifies the resource for which the token is requested.
+     * @returns {string} token if if it exists and not expired, otherwise null.
      */
     AuthenticationContext.prototype.getCachedToken = function (resource) {
         if (!this._hasResource(resource)) {
@@ -303,8 +320,16 @@ var AuthenticationContext = (function () {
             return null;
         }
     };
+
     /**
-     * Retrieves and parse idToken from localstorage
+    * User information from idtoken.
+    *  @class User
+    *  @property {string} userName - username assigned from upn or email.
+    *  @property {object} profile - properties parsed from idtoken.
+    */
+
+    /**
+     * If user object exists, returns it. Else creates a new user object by decoding id_token from the cache.
      * @returns {User} user object
      */
     AuthenticationContext.prototype.getCachedUser = function () {
@@ -316,7 +341,13 @@ var AuthenticationContext = (function () {
         this._user = this._createUser(idtoken);
         return this._user;
     };
-
+  
+    /**
+     * Adds the passed callback to the array of callbacks for the specified resource and puts the array on the window object. 
+     * @param {string}   resource A URI that identifies the resource for which the token is requested.
+     * @param {string}   expectedState A unique identifier (guid).
+     * @param {tokenCallback} callback - The callback provided by the caller. It will be called with token or error.
+     */
     AuthenticationContext.prototype.registerCallback = function (expectedState, resource, callback) {
         this._activeRenewals[resource] = expectedState;
         if (!window.callBacksMappedToRenewStates[expectedState]) {
@@ -347,8 +378,7 @@ var AuthenticationContext = (function () {
     // with callback
     /**
      * Acquires access token with hidden iframe
-     * @param {string}   resource  ResourceUri identifying the target resource
-     * @returns {string} access token if request is successful
+     * @ignore
      */
     AuthenticationContext.prototype._renewToken = function (resource, callback) {
         // use iframe to try refresh token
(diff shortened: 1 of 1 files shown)


lib/adal.js
```diff
@@ -22,29 +22,23 @@ var AuthenticationContext = (function () {
...
-    /**
-     * User information from idtoken.
-     *  @class User
-     *  @property {string} userName - username assigned from upn or email.
-     *  @property {object} profile - properties parsed from idtoken.
+     * Configuration options for Authentication Context.
+     * @class config
+     *  @property {string} tenant - Your target tenant.
+     *  @property {string} clientID - Client ID assigned to your app by Azure Active Directory.
+     *  @property {string} redirectUri - Endpoint at which you expect to receive tokens. Set this value only if you want to redirect user to a page other than the login start page after successful authentication.
```
tushargupta51: The second half is not correct. Even if this is set, user will be redirected to login start page. Also, specify the default values wherever applicable. For example: instance, tenant, redirectUri, popUp.

lib/adal.js
```diff
@@ -22,29 +22,23 @@ var AuthenticationContext = (function () {
...
+     *  @property {string} instance - Azure Active Directory Instance (default:https://login.microsoftonline.com/).
+     *  @property {Array} endpoints - Collection of {Endpoint-ResourceId} used for autmatically attaching tokens in webApi calls.
```
tushargupta51: *automatically

lib/adal.js
```diff
@@ -270,17 +272,26 @@ var AuthenticationContext = (function () {
...
     AuthenticationContext.prototype._hasResource = function (key) {
         var keys = this._getItem(this.CONSTANTS.STORAGE.TOKEN_KEYS);
         return keys && !this._isEmpty(keys) && (keys.indexOf(key + this.CONSTANTS.RESOURCE_DELIMETER) > -1);
     };
 
     /**
-     * Gets token for the specified resource from local storage cache
+     * Gets token for the specified resource from the cache.
      * @param {string}   resource A URI that identifies the resource for which the token is valid.
      * @returns {string} token if exists and not expired or null
```
tushargupta51: can we update this to: token, if it exists and not expired, otherwise null

lib/adal.js
```diff
@@ -317,6 +336,11 @@ var AuthenticationContext = (function () {
         return this._user;
     };
 
+    /**
+     * If multiple requests for acquireToken are made for the same resource, we send only the first request and register the callbacks for all subsequent requests.
+     * When we process the response from AAD, this method calls all the callbacks with the error/token.
```
tushargupta51: This should describe what this method does. All it does is add the passed callback to the array of callbacks for the specified resource and put the array on the window object. It does not call the callbacks itself.

lib/adal.js
```diff
@@ -438,9 +476,9 @@ var AuthenticationContext = (function () {
     };
 
     /**
-     * Acquire token from cache if not expired and available. Acquires token from iframe if expired.
+     * Acquires token from cache if not expired and available. Acquires token from iframe if expired.
      * @param {string}   resource  ResourceUri identifying the target resource
-     * @param {requestCallback} callback
+     * @param {requestCallback} callback -  The callback that contains the token or error
```
tushargupta51: *The callback provided by the caller. It will be called with token or error.

lib/adal.js
```diff
@@ -438,9 +476,9 @@ var AuthenticationContext = (function () {
     };
 
     /**
-     * Acquire token from cache if not expired and available. Acquires token from iframe if expired.
+     * Acquires token from cache if not expired and available. Acquires token from iframe if expired.
```
tushargupta51: Acquires token from cache if there is a valid token present in the cache. Otherwise sends token request to AAD for a new token.

lib/adal.js
```diff
@@ -481,8 +519,8 @@ var AuthenticationContext = (function () {
     };
 
     /**
-     * Redirect the Browser to Azure AD Authorization endpoint
-     * @param {string}   urlNavigate The authorization request url
+     * Redirect the browser to Azure AD authorization endpoint
+     * @ignore
```
tushargupta51: this is a public method

lib/adal.js
```diff
@@ -494,7 +532,8 @@ var AuthenticationContext = (function () {
     };
 
     /**
-     * Clear cache items.
+     * Clears cache items.
+     * @ignore
```
tushargupta51: public method

lib/adal.js
```diff
@@ -519,7 +558,8 @@ var AuthenticationContext = (function () {
     };
 
     /**
-     * Clear cache items for a resource.
+     * Clears cache items for a given resource.
+     * @ignore
```
tushargupta51: public method

lib/adal.js
```diff
@@ -532,8 +572,8 @@ var AuthenticationContext = (function () {
     };
 
     /**
-     * Logout user will redirect page to logout endpoint.
-     * After logout, it will redirect to post_logout page if provided.
+     * Redirects user to logout endpoint.
+     * After logout, it will redirect to post_logout page if specified.
```
tushargupta51: post_logout -> post_logout_redirect_uri (or whatever is the name of the property)

lib/adal.js
```diff
@@ -560,13 +600,14 @@ var AuthenticationContext = (function () {
...
+     * @param {string} error error message if user info is not available
+     * @param {User} user user object retrieved from the cache
      */
 
     /**
-     * Gets a user profile
-     * @param {requestCallback} cb - The callback that handles the response.
+     * Calls the passed in callback with the user object or error message related to the user.
+     * @param {requestCallback} callback - The callback that handles the response.
+     * @returns {User} user object.
```
tushargupta51: It does not return the user object

lib/adal.js
```diff
@@ -662,15 +716,33 @@ var AuthenticationContext = (function () {
...
      */
     AuthenticationContext.prototype.getLoginError = function () {
         return this._getItem(this.CONSTANTS.STORAGE.LOGIN_ERROR);
     };
 
     /**
-     * Gets requestInfo from given hash.
-     * @returns {string} error message related to login
+     * Parameters object created from the response received from AAD.
+     *  @class Parameters
```
tushargupta51: I don't think this qualifies for a class. This is the list of query parameters passed in the response from AAD. We cannot really define the interface for this as it can change based on query parameters in the response.

lib/adal.js
```diff
@@ -662,15 +716,33 @@ var AuthenticationContext = (function () {
...
+     *  @property {string} id_token - id_token of the user.
+     *  @property {string} session_state - state identifying the user's session.
+     *  @property {string} state - unique guid used to match the response with the request.
+     */
+
+    /**
+     * Request info object created from the response received from AAD.
+     *  @class RequestInfo
+     *  @property {Parameters} parameters - object comprising of id_token/error, session_state and state.
```
tushargupta51: Instead of {Parameters}, you can use something like dictionary here.

lib/adal.js
```diff
@@ -662,15 +716,33 @@ var AuthenticationContext = (function () {
...
+     *  @property {requestType} requestType - either LOGIN, RENEW_TOKEN or UNKNOWN.
```
tushargupta51: you can link the enum type here for REQUEST_TYPE.

lib/adal.js
```diff
@@ -739,9 +815,7 @@ var AuthenticationContext = (function () {
     };
 
     /**
-     * Saves token from hash that is received from redirect.
-     * @param {string} hash  -  Hash passed from redirect page
-     * @returns {string} error message related to login
+     * Saves token from hash that is received from the redirect response from AAD.
```
tushargupta51: saves token or error in the cache received in the response from AAD. In case of id_token, it also creates the user object.

lib/adal.js
```diff
@@ -909,6 +994,10 @@ var AuthenticationContext = (function () {
         return urlNavigate;
     };
 
+    /**
+     * Extracts the id token.
```
tushargupta51: Returns the decoded id_token.

lib/adal.js
```diff
@@ -1015,10 +1112,18 @@ var AuthenticationContext = (function () {
...
+    /**
+     * Converts string to represent binary data in ASCII string format by translating it into a radix-64 representation and returns it
+     * @ignore
+     */
     AuthenticationContext.prototype._convertUrlSafeToRegularBase64EncodedString = function (str) {
         return str.replace('-', '+').replace('_', '/');
     };
 
+    /**
+     * Generates the parameters for the authorization endpoint URL.
```
tushargupta51: Serializes the parameters for the authorization endpoint URL and returns the serialized uri string.

lib/adal.js
```diff
@@ -1132,15 +1249,26 @@ var AuthenticationContext = (function () {
...
+    /**
+     * Calculates the expires in value in milliseconds for the acquired token
+     * @ignore
+     */
     AuthenticationContext.prototype._expiresIn = function (expires) {
         return this._now() + parseInt(expires, 10);
     };
 
+    /**
+     * Returns the number of milliseconds
```
tushargupta51: Returns the current time as number of milliseconds since ___ (not sure what this value is)

lib/adal.js
```diff
@@ -1232,6 +1376,10 @@ var AuthenticationContext = (function () {
         }
     };
 
+    /**
+     * Creates a copy of the config object passed to the constructor.
```
tushargupta51: The description should only describe what this method does. You can write a generic description for this: returns cloned copy of the passed object, or something along those lines

lib/adal.js
```diff
@@ -1246,12 +1394,20 @@ var AuthenticationContext = (function () {
...
+     */
     AuthenticationContext.prototype._addLibMetadata = function () {
         // x-client-SKU
         // x-client-Ver
         return '&x-client-SKU=Js&x-client-Ver=' + this._libVersion();
     };
 
+    /**
+     * Checks the Logging Level and constructs the Log message.
+     * @ignore
```
tushargupta51: public method. Besides constructing the log message, it logs the message also. In the description, we should specify that users should implement/override this method to turn on logging.

lib/adal.js
```diff
@@ -1270,26 +1426,50 @@ var AuthenticationContext = (function () {
         }
     };
 
+    /**
+     * Logs messages to the console window of the browser when Logging Level is set to 0.
+     * @ignore
```
tushargupta51: this and all the other log methods are public.

lib/adal.js
```diff
@@ -335,11 +335,17 @@ var AuthenticationContext = (function () {
         this._user = this._createUser(idtoken);
         return this._user;
     };
-
     /**
-     * If multiple requests for acquireToken are made for the same resource, we send only the first request and register the callbacks for all subsequent requests.
-     * When we process the response from AAD, this method calls all the callbacks with the error/token.
-     * @returns {User} user object
+     * This callback is displayed as part of the Requester class.
```
tushargupta51: This does not look like the right description.

lib/adal.js
```diff
@@ -520,7 +526,6 @@ var AuthenticationContext = (function () {
 
     /**
      * Redirect the browser to Azure AD authorization endpoint
```
tushargupta51: Add description for the parameter

lib/adal.js
```diff
@@ -559,7 +563,7 @@ var AuthenticationContext = (function () {
 
     /**
      * Clears cache items for a given resource.
-     * @ignore
+     * @param {string}  resource A URI that identifies the resource for which the token is requested.
```
tushargupta51: "token is requested" part is not correct here.

lib/adal.js
```diff
@@ -573,7 +577,7 @@ var AuthenticationContext = (function () {
 
     /**
      * Redirects user to logout endpoint.
-     * After logout, it will redirect to post_logout page if specified.
+     * After logout, it will redirect to postLogoutRedirectUri if added as a property on the config object.Defaults to 'window.location.href'.
```
tushargupta51: There is no default for this at adal level.

lib/adal.js
```diff
@@ -335,11 +335,17 @@ var AuthenticationContext = (function () {
         this._user = this._createUser(idtoken);
         return this._user;
     };
-
     /**
-     * If multiple requests for acquireToken are made for the same resource, we send only the first request and register the callbacks for all subsequent requests.
-     * When we process the response from AAD, this method calls all the callbacks with the error/token.
-     * @returns {User} user object
+     * This callback is displayed as part of the Requester class.
+     * @callback requestUserCallback
```
tushargupta51: You declared this class here but I don't see it being used here. Callbacks for both `acquiretoken` and `getUser` is same: requestCallback. I think you should define 2 classes: tokenCallback and userCallback. Set the callbacks for token related methods to be of type tokenCallback and for user related, set it to userCallback

lib/adal.js
```diff
@@ -482,9 +477,15 @@ var AuthenticationContext = (function () {
     };
 
     /**
+     * @callback tokenCallback
+     * @param {string} error error message returned from AAD if silent token request fails.
+     * @param {string} token token returned from AAD if silent token request is successful.
```
tushargupta51: We should remove the 'silent' word, as the same callback will be mostly be used for interactive flows too, whenever we add them.

lib/adal.js
```diff
@@ -22,29 +22,23 @@ var AuthenticationContext = (function () {
...
-     * @property {endpoints}       Collection of {Endpoint-ResourceId} used for autmatically attaching tokens in webApi calls
-     */
-
-    /**
-     * User information from idtoken.
-     *  @class User
-     *  @property {string} userName - username assigned from upn or email.
-     *  @property {object} profile - properties parsed from idtoken.
+     * Configuration options for Authentication Context.
+     * @class config
```
tushargupta51: We are missing some public properties on this class: callback, displayCall, anonymousEndpoints, expireOffsetSeconds, logOutUri, postLogoutRedirectUri, cacheLocation, correlationId. We can skip `isAngular` as we don't want users to use it for plain JS.

lib/adal.js
```diff
@@ -559,14 +605,14 @@ var AuthenticationContext = (function () {
 
     /**
      * This callback is displayed as part of the Requester class.
```
tushargupta51: We should remove this.

lib/adal.js
```diff
@@ -822,8 +893,8 @@ var AuthenticationContext = (function () {
 
     /**
      * Gets resource for given endpoint if mapping is provided with config.
-     * @param {string} endpoint  -  API endoibt
-     * @returns {string} resource for this API endpoint
+     * @param {string} endpoint  -  API endpoint
```
tushargupta51: Can we change the description to: The uri for which the resource Id is requested.

lib/adal.js
```diff
@@ -169,8 +163,7 @@ var AuthenticationContext = (function () {
     };
 
     /**
-     * Gets initial Idtoken for the app backend
-     * Saves the resulting Idtoken in localStorage.
+     * Gets initial Idtoken for the app's backend and saves it in the cache.
```
tushargupta51: We should update this to something like: "Initiates the login process, redirecting or prompting the user to Azure AD login page".

lib/adal.js
```diff
@@ -870,7 +945,10 @@ var AuthenticationContext = (function () {
         return extractedUri;
     };
 
-    /*exported  oauth2Callback */
+    /**
+     * This method must be called for processing the response received from AAD. It extracts the hash, processes the token or error, saves it in the cache and calls the registered callbacks with the result.
+     * @ignore
```
tushargupta51: Re-commenting:
public method. This is the main method for processing responses received from AAD for users using plain JS. This MUST be called in order to process the response. This can say something like:
This method must be called for processing the response received from AAD. It extracts the hash, process the token or error, saves it in the cache and calls the registered callbacks with the result.

tushargupta51:
:shipit:

PR #437, opened 2016-11-15
save 500ms from each token aquisition

kirkshoop:
also fix the .src check

Credit to Gevorg Kurghinyan and Rajesh Babu Srikakollu for these changes

### DIFF
--- a/lib/adal.js
+++ b/lib/adal.js
@@ -1,4 +1,4 @@
-//----------------------------------------------------------------------
+﻿//----------------------------------------------------------------------
 // AdalJS v1.0.13
 // @preserve Copyright (c) Microsoft Open Technologies, Inc.
 // All Rights Reserved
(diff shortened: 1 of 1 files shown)


lib/adal.js
```diff
@@ -475,11 +475,11 @@ var AuthenticationContext = (function () {
         var frameCheck = frameName;
         setTimeout(function () {
             var frameHandle = self._addAdalFrame(frameCheck);
-            if (frameHandle.src === '' || frameHandle.src === 'about:blank') {
```
kirkshoop, line 478: Gevorg: "When you add a new iframe to the document,in IE the 'src' attribute of the iframe is set to 'undefined', but the code handles only empty string and 'blank'."

lib/adal.js
```diff
@@ -475,11 +475,11 @@ var AuthenticationContext = (function () {
...
+            if (!frameHandle.src || frameHandle.src === 'about:blank') {
                 frameHandle.src = urlNavigate;
                 self._loadFrame(urlNavigate, frameCheck);
             }
-        }, 500);
+        }, 0);
```
kirkshoop: Kirk: "The failure in IE is not a race. It is a reentrancy issue. Just letting the stack unwind and then immediately setting the .src works."

lib/adal.js
stevegomez17: Believe it’s cause it’s a null api value needs to be 0

PR #467, opened 2017-01-20
Rename variable for clarity

magnuf:
Make it more clear that the variable is the expiration time, and not if it is expired or not.


tushargupta51:
Thanks @magnuf for your contribution!

Issue #462, opened 2016-12-20
Call displayCall method before using any default behavior in login method


AidaNow:
Hi,

I am trying to add '&prompt=admin_consent' to 'urlNavigate' to be able to get the IT admin consent. I am using the following code suggested [here](https://github.com/waldekmastykarz/sample-azure-adminconsent/blob/master/app/admin/admin.controller.js#L4-L5):

var adal = new AuthenticationContext();
adal.config.displayCall = function adminFlowDisplayCall(urlNavigate) {
urlNavigate += '&prompt=admin_consent';
adal.promptUser(urlNavigate);
};
adal.login();
adal.config.displayCall = null;

but it does not work for me because I am using the 'popup: true' flag when I initialize the adal library (I am using version 1.0.13). The reason is that in adal.js, we have the following block of code:

    /**
     * Initiates the login process by redirecting the user to Azure AD authorization endpoint.
     */
    AuthenticationContext.prototype.login = function () {
        // Token is not present and user needs to login
        if (this._loginInProgress) {
            this.info("Login in progress");
            return;
        }
        var expectedState = this._guid();
        this.config.state = expectedState;
        this._idTokenNonce = this._guid();
        this.verbose('Expected state: ' + expectedState + ' startPage:' + window.location);
        this._saveItem(this.CONSTANTS.STORAGE.LOGIN_REQUEST, window.location);
        this._saveItem(this.CONSTANTS.STORAGE.LOGIN_ERROR, '');
        this._saveItem(this.CONSTANTS.STORAGE.STATE_LOGIN, expectedState);
        this._saveItem(this.CONSTANTS.STORAGE.NONCE_IDTOKEN, this._idTokenNonce);
        this._saveItem(this.CONSTANTS.STORAGE.ERROR, '');
        this._saveItem(this.CONSTANTS.STORAGE.ERROR_DESCRIPTION, '');
        var urlNavigate = this._getNavigateUrl('id_token', null) + '&nonce=' + encodeURIComponent(this._idTokenNonce);
        this._loginInProgress = true;
        if (this.popUp) {
            this._loginPopup(urlNavigate);
            return;
        }
        if (this.config.displayCall) {
            // User defined way of handling the navigation
            this.config.displayCall(urlNavigate);
        } else {
            this.promptUser(urlNavigate);
        }
    };
	
[...]

PR #477, opened 2017-02-05
call displayCall before any default behavior in login method

rohitnarula7176:
#462


tushargupta51:
:shipit:  you might need to rebase it from before merging since the changes in test code were in an another PR too.

Issue #411, opened 2016-10-05
Intermittently seeing expiration key off by 4 minutes


stuartlong:
We are hitting an intermittent issue in our application where the expiration time stored in **adal.expiration.key<resource>** is off by a number of minutes. Our app authenticates clientside with ADAL.js and sends the token down to our backend service when we make service calls. Our backend service then reauthenticates the token. 

When we hit the issue, we make an **acquireToken** call to ADAL.js, which returns the cached token because it thinks it's not expired, and send that token down to the service. However, the service request gets refused because when it does the authentication, the token is expired. I grabbed one of the tokens that was causing this issue, and, indeed, it's **exp** value was different than that set for **adal.expiration.key**.

tushargupta51:
@stuartlong This might be happening because of clock skews between server and client. You can set `.config.expireOffsetSeconds` property to 4 minutes or so. Default is 2 minutes. What this property does is that if the cached token is about to be expired in the set time interval, it will renew the token instead of using the cached token.

stuartlong:
@tushargupta51 I don't believe that's the case. First, we checked the clocks on the machines. They were off by a couple of seconds, but not 4 minutes. Second, that wouldn't explain the fact that when I decoded the client machine's token, it's **exp** value was different than that set for **adal.expiration.key**. In fact, it's **exp** value was expired to 4 minutes before we sent the service request.

tushargupta51:
@stuartlong For access token, we use the "expires in" value returned as query parameter. The value by default is 3600 seconds == 1 hour. We then calculate the number of seconds since 1970 and add 3600 to that and save as expiration time of the token. The fact that you are seeing difference between "exp" and "adal.expiration.key", that indicates to me that there is a clock difference between the client where JS code is running and the server.

stuartlong:
@tushargupta51 Ah, you mean the AAD server that's actually returning the token, not our backend severs?  Also, why don't you consider the "exp" value of the returned token? Wouldn't doing so avoid issues like this?

tushargupta51:
@stuartlong I actually meant both. It could be that your backend server is out of sync, and thus throwing error when validating the token. But you mentioned that you checked that. The fact that the "exp" and "adal.expiration.key" values differ seems to indicate that the client clock and Azure AD clock are not in sync.

I agree that we can use the "exp" value of an access token like we do for id_token. Also, as per spec, "expires_in" is an recommended but optional property. We will fix this issue.

[...]

PR #483, opened 2017-02-08
update the expires offset from 2 to 5 minutes, use default value for …

tushargupta51:
…"expires_in"

#411


rohitnarula7176:
:shipit:

PR #498, opened 2017-02-18
pass event from locationChangeHandler to processhash

lukas-gmuer:
the event variable is not defined anymore in the processhash function and therefore the call to event.preventDefault fails and the following callback will never be executed.

### DIFF
--- a/lib/adal-angular.js
+++ b/lib/adal-angular.js
@@ -75,7 +75,7 @@
                     else {
                         var hash = '#' + $location.path();
                     }
-                    processHash(hash);
+                    processHash(hash, event);
 
                     $timeout(function () {
                         updateDataFromCache(_adal.config.loginResource);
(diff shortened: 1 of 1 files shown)


tushargupta51:
@lukas-gmuer Thanks for the PR. We do need to fix this. I will review the PR.

rohitnarula7176:
@lukas-gmuer Can you explain the scenario in detail when you see this event object being undefined. Even if we don't pass this event to the processHash function, it takes the form of the event "DomContentLoaded" as it is globally defined. Any information regarding the repro steps or the version of angular and angular-route would be helpful.

lukas-gmuer:
angular and angular-route both 1.5.8

It happens in all cases where we come from the locationChangeHandler. 
I guess these are the scenarios for getting a token to another resource(endpoint) or a token renewal.

The event is not globally defined if we come from locationChangeHandler. 
As that is not an eventlistener for the window, but an angular event listener.

PR #505, opened 2017-02-27
Only redirect to original url in case of login req

cedricboon:
Redirection to the original page that was saved to storage before initiating the login should only happen when handling a login request. If the target page performs API calls, and the redirect is also performed when handling token renewals, this results in decreased performance due to the API calls being fired multiple times.


rohitnarula7176:
@cedricboon Thank you for your contribution but we need the redirection back to the start page for both login as well as when you are trying to acquire a token interactively instead of a hidden iframe.

Issue #514, opened 2017-03-07
Broken link in main README.md file


valera-rozuvan:
The link of **Claims in Azure AD Security Tokens** in file [README.md](https://github.com/AzureAD/azure-activedirectory-library-for-js/blob/dev/README.md) points to [https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-authentication-scenarios%23claims-in-azure-ad-security-tokens](https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-authentication-scenarios%23claims-in-azure-ad-security-tokens). When you navigate there, the site reports an error.

rohitnarula7176:
Fixed.

PR #515, opened 2017-03-07
Fix broken link

valera-rozuvan:
Fix for issue https://github.com/AzureAD/azure-activedirectory-library-for-js/issues/514 .


FrancescoRizzi:
The link in the commit is to the correct page, but I believe it's missing the anchor to the Claims section ( [#claims-in-azure-ad-security-tokens](https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-authentication-scenarios#claims-in-azure-ad-security-tokens) ). HTH

valera-rozuvan:
@FrancescoRizzi It seems that GitHub markdown does something wrong with the symbol `#` in URLs. That's why the URL come out broken when rendered as HTML...

tushargupta51:
Thanks @valera-rozuvan  for your contribution. PR looks good to me. I will merge it.

PR #520, opened 2017-03-14
Add domain_hint optional feature to login

brentos99:
Added a Config option "domainHint" that will allow provide the domain_hint optional parameter when initially authenticating via ADAL. 

> This value can be consumers or organizations. If included, it skips the email-based discovery process that the user goes through on the v2.0 sign-in page, for a slightly more streamlined user experience.

https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-v2-protocols-oidc

to use the feature via the config 

adalAuthenticationServiceProvider.init(
        {
            // Config to specify endpoints and similar for your app
            tenant: "52d4b072-9470-49fb-8721-bc3a1c9912a1", // Optional by default, it sends common
            clientId: "e9a5a8b6-8af7-4719-9821-0deef255f68e", // Required
            //localLoginUrl: "/login",  // optional
            //redirectUri : "your site", optional
            domainHint: "organizations" // optional, value can be consumers or organizations
            
        },


tushargupta51:
@brentos99 We have `extraQueryParameters` for that purpose. Users can set domain_hint like

`extraQueryParameters: "domain_hint=organizations"

brentos99:
Where do you set that?

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

PR #542, opened 2017-04-21
Ability to prevent automatic token injection through $http config flag

r4hulp:
There can be a scenario where an application requires routes to be guarded for authorized users. 

However, some of the outgoing requests do require special bearer token along with requests. To avoid broken requests in such scenarios set **`preventAccessTokenInjection:true`** in configuration of **`$http`** request. This will prevent adal-angular from auto injecting access token in that particular request. 

Some example snippets given below - 

```js
$http({
  method: 'GET',
  url: '/someUrl',
  preventAccessTokenInjection : true
}).then(function successCallback(response) {
    // this callback will be called asynchronously
    // when the response is available
  }, function errorCallback(response) {
    // called asynchronously if an error occurs
    // or server returns response with an error status.
  });
```

```js
$http.get('/someUrl', {
        preventAccessTokenInjection : true
    }).then(successCallback, errorCallback);
```


rahulrulez:
The least thing I am expecting is repository maintainers to get involve in conversation at least :-) @rohitnarula7176 @tushargupta51

PR #547, opened 2017-04-29
Don't enforce user info in acquireToken()

zeevm:
user info/idToken isn't necessary to obtain access tokens in implicit grant flow, adal should allow acquiring tokens purely based on the existence of AAD session cookie.


rohitnarula7176:
@zeevm As discussed, we cannot make the user check configurable as per design of the library. Closing the pull request. Thank you for your contribution.

Issue #536, opened 2017-04-06
ui-router's ui-sref + adal = return to the page your on.. not the page your going to..


jhandel:
I seem to have found an issue.. If I click on a link using ui-sref for a page that requires authentication then after authentication I return to the page I started at (not the page I want to go to).. I have confirmed that the issue is that adal.login.request is not storing my target page but my origination page.

If I manually type the URL in for the authenticated page then I do go to that page after login.. So I am pretty sure it has to how state is being handled vs. what ever step stores adal.login.request in the login pipeline.

Thanks
Josh

rohitnarula7176:
@jhandel Thank you for pointing out this issue. It is now fixed in dev.

PR #548, opened 2017-05-02
Use url generated from target state for adal.login.request

ccurrens:
This is a suggested fix for #536. The main issue is that `$location` is used for determining the return URL after login. However, the `$stateChangeStart` event doesn't update the `$location` url, so the user is returned back to the original state instead of the protected one they were trying to get to.

This change generates the target URL from the state inside of the `$stateChangeStart` event, and modifies the login handler to accept the generated `returnUrl` as an an optional parameter.

### DIFF
--- a/lib/adal-angular.js
+++ b/lib/adal-angular.js
@@ -179,7 +179,7 @@
 
                 };
 
-                var loginHandler = function () {
+                var loginHandler = function (returnUrl) {
                     _adal.info('Login event for:' + $location.$$url);
                     if (_adal.config && _adal.config.localLoginUrl) {
                         $location.path(_adal.config.localLoginUrl);
@@ -188,7 +188,7 @@
                         // directly start login flow
                         _adal.info('Start login at:' + $location.$$absUrl);
                         $rootScope.$broadcast('adal:loginRedirect');
-                        _adal.login($location.$$absUrl);
+                        _adal.login(returnUrl || $location.$$absUrl);
                     }
                 };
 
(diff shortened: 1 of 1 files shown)


lib/adal-angular.js
```diff
@@ -264,7 +264,8 @@
                                 if (!_oauthData.isAuthenticated) {
                                     if (!_adal._renewActive && !_adal.loginInProgress()) {
                                         _adal.info('State change event for:' + $location.$$url);
-                                        loginHandler();
+                                        var $state = $injector.get('$state');
+                                        loginHandler(window.location.origin + '/' + $state.href(toState, toParams));
```
rohitnarula7176: We do not need to use the window.origin property here. Please change this to:  
var loginStartPage = $state.href(toState, toParams, { absolute: true })
loginHandler(loginStartPage);

lib/adal-angular.js
```diff
@@ -179,7 +179,7 @@
 
                 };
 
-                var loginHandler = function () {
+                var loginHandler = function (returnUrl) {
                     _adal.info('Login event for:' + $location.$$url);
```
rohitnarula7176: if loginStartPage is passed in the function, the _adal.info should print the loginStartPage instead of the absolute url used earlier. Please change this to:
var loginUrl = loginStartPage || $location.$$url;
                    if (_adal.config && _adal.config.localLoginUrl) {
                        _adal.info('Login event for:' + _adal.config.localLoginUrl);
                        $location.path(_adal.config.localLoginUrl);
                    }
                    else {
                        // directly start login flow
                        _adal.info('Start login at:' + loginUrl);
                        $rootScope.$broadcast('adal:loginRedirect');
                        _adal.login(loginUrl);
                    }

ccurrens:
I see this was fixed with #637 , so I'll close it.

Issue #543, opened 2017-04-24
adal.js not refreshing token properly 


bllfrnch:
Hi,

I wrote up a question on Stack Overflow, too, but I'm wondering if this may be a bug, so I'm going to report it.

I'm working on a SPA written in Vue.js. The site is running on one host, web.xxxx.com, while the api is running on another, api.xxxx.com. API calls are secured, and we use the low level adal.js library for login, logout, and also to provide the logged in user's token to the API. In Vue, we set up an interceptor as follows (`getToken` is a wrapper around `acquireToken` that returns a promise):

    Vue.http.interceptors.push((request, next) => {
      // if this is a call to API, then set up auth
      Vue.http.options.xhr = {withCredentials: true};
      getToken()
        .then((token) => {
          request.headers.set('Authorization', `Bearer ${token}`);
          next();
        })
        .catch((err) => {
          throw new Error(err);
        });
    });

This works as expected; whenever an HTTP request is initiated, the interceptor sets additional headers and off it goes.

However, adal.js isn't updating the tokens. After an hour, my logged in user is not longer logged in, despite interacting with the API. I've even tried running this, to ensure the token isn't being refreshed:

    // ping every minute...
    window.setInterval(() => {
      getToken().then((token) => { window.Logging.log('GOT IT!!!', token); });
    }, 60 * 1000);

After 55 minutes, these entries appear in the client logs. I redacted anything that may be sensitive, and it should be noted that this takes so long because I was manually stepping through all the calls in the debugger:

    Fri, 21 Apr 2017 22:11:32 GMT:1.0.14-VERBOSE: renewing idtoken
    Fri, 21 Apr 2017 22:11:45 GMT:1.0.14-INFO: renewIdToken is called
    Fri, 21 Apr 2017 22:12:08 GMT:1.0.14-INFO: Add adal frame to document:adalIdTokenFrame
    Fri, 21 Apr 2017 22:13:26 GMT:1.0.14-VERBOSE: Renew Idtoken Expected state: <REDACTED>
    Fri, 21 Apr 2017 22:13:36 GMT:1.0.14-INFO: Navigate url:https://login.microsoftonline.com/<REDACTED>&x-client-SKU=Js&x-client-Ver=1.0.14
    Fri, 21 Apr 2017 22:14:06 GMT:1.0.14-VERBOSE: Navigate to:https://login.microsoftonline.com/<REDACTED>
[...]

PR #552, opened 2017-05-12
Added renewStates to window

### LINKED ISSUE #460: Importing adal.js as a module breaks token refresh
There is some module loading code in adal.js, so I tried bundling it with webpack and "require"ing it.  everything seemed to work fine till I went to refresh and the code at https://github.com/AzureAD/azure-activedirectory-library-for-js/blob/dev/lib/adal.js#L796 tried to see if the IFrame window.parent.AuthenticationContext was defined.  When importing as a module that property isn't defined.  At that point the rest of the code thinks that the state didn't match.

rohitnarula7176:
Fixes #460 #543

### DIFF
--- a/lib/adal.js
+++ b/lib/adal.js
@@ -119,8 +119,8 @@ var AuthenticationContext = (function () {
         this._user = null;
         this._activeRenewals = {};
         this._loginInProgress = false;
-        this._renewStates = [];
 
+        window.renewStates = [];
         window.callBackMappedToRenewStates = {};
         window.callBacksMappedToRenewStates = {};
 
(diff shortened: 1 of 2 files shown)


lovemaths:
:shipit:

nybroe:
Have these changes been removed again?
I can see that in the latest version you are still adding renewStates to the current context (this) and not the window as the changes in this PR states.

rohitnarula7176:
@nybroe The name of the PR is misleading. We only changed the logic once where we moved properties renewStates, callBackMappedToRenewStates and callBacksMappedToRenewStates from window to the private Adal instance as they were causing problems when consumed in webpack in the case when the module name was defined as AuthenticationContext (the same name Adal exports as global on the window)

nybroe:
@rohitnarula7176 Thank you for the quick response. Even though all the properties are moved to the adal instance, I can still see that _renewStates and others are empty or undefined when read through the iframe.
Please take a look at my newly created issue. :)
[https://github.com/AzureAD/azure-activedirectory-library-for-js/issues/710](url)

Issue #578, opened 2017-06-22
Token Renewal Failed | Token renewal operation failed due to timeout


balajimadgundi:
I am using the Adal 1.14.0 version and sessionStorage. Below is App init:
adalProvider.init({
                tenant: Dependencies.Tenant,
                clientId: Dependencies.ClientId,
                endpoints: Dependencies.Endpoints,
                postLogoutRedirectUri: Dependencies.ClientGatewayLogoutUrl
            }, $httpProvider);

I am getting the "Token renewal operation failed due to timeout" issue when trying to open the application in new Tab/window of IE browser(IE 11) for subsequent login.

When I am logging in first time(first tab) it works. For subsequent login in new Tab/Window in IE, I am getting this "Token renewal operation failed due to timeout" from AcquireToken:Failure event.

I have checked this on other users machines(4-5), The same scenario is working fine on other machine's IE browser. Does not work only in my machine's IE subsequent login scenario.

I have already verified all my IE settings, they are same as of other user's IE settings where it is working. I need to know if this is causing due to any IE settings or anything else. I can not rollout the code just because it is working in 3-4 machines since it is failing in my machine and other application user base might have same IE settings as mine(if this is causing due to IE settings).

PFA logs
[SessionStorageIssueLogs.txt](https://github.com/AzureAD/azure-activedirectory-library-for-js/files/1096199/SessionStorageIssueLogs.txt)

rohitnarula7176:
@balajimadgundi Can you share a working demo of your code so I can debug this further. Also can you please attach the fiddler trace for this issue so I can confirm that you actually receive a token from AAD.

thdotnet:
I was facing the same problem. Here's how I've solved it:

    /**
     * Gets token for the specified resource from the cache.
     * @param {string}   resource A URI that identifies the resource for which the token is requested.
     * @returns {string} token if if it exists and not expired, otherwise null.
     */
    AuthenticationContext.prototype.getCachedToken = function (resource) {
        if (!this._hasResource(resource)) {
            return null;
        }

        var token = this._getItem(this.CONSTANTS.STORAGE.ACCESS_TOKEN_KEY + resource);
        var expiry = this._getItem(this.CONSTANTS.STORAGE.EXPIRATION_KEY + resource);

        // If expiration is within offset, it will force renew
        var offset = this.config.expireOffsetSeconds || 300;

        if (expiry && (expiry > this._now() + offset)) {
            return token;
        } else {
[...]

PR #581, opened 2017-06-28
Token Renewal Failed

thdotnet:
Fixing issue #578

I was able to reproduce the issue #578 using just adal.js in a React SPA. To solve, I've removed from the sessionStorage the keys related to the adal. After that, it's working.


rohitnarula7176:
@thdotnet Thank you for your contribution but the token renewal operation cannot fail because of the presence of a blank access token and expiration key in the storage. Please open an issue if you still encounter any problems and we will be happy to have a look at it.

PR #598, opened 2017-07-20
Hide iframe contents from screen reader

rcasto:
Ran into a case at least with VoiceOver on iOS and Mac, where the iframe contents for adal were being read and getting focus by screen reader.  It may also affect other browser + screen reader combinations. This interrupts focus order/navigation for these users and is an unpleasant experience in general.


rohitnarula7176:
@rcasto Thank you for your contribution.

Issue #574, opened 2017-06-16
adal.js doesn't works with ui-router release 1


mathilde-pellerin:
I spent few hours to try to have ADAL working with my app but even though my configuration seems to be good, the redirection to the authentication page is never triggered. 
So I took a look into adal-angular.js and I found that it relies on events which are not triggered by ui-router release 1:

```javascript
                // Route change event tracking to receive fragment and also auto renew tokens
                $rootScope.$on('$routeChangeStart', routeChangeHandler);

                $rootScope.$on('$stateChangeStart', stateChangeHandler);

                $rootScope.$on('$locationChangeStart', locationChangeHandler);

                $rootScope.$on('$stateChangeError', stateChangeErrorHandler);
```

UI-router release 1 use transition hooks now. That would be great to have a version of adal-angular.js compatible with UI-router 1 now it has been released.

gfsarri:
I'm also having problems between ui-router (old version, < 1.x)  and ADAL.  In my case what I can see is that there must be some kind of issue between the ui-router navigation and the attempt to request Azure the navigation token (window.location.replace(urlNavigate); inside AuthenticationContext.prototype.promptUser in adal.js).

In case it can help somebody, the following **workaround** has worked for me: set to false requireADLogin in the ui-router state and, after the state has loaded, call explicitly from the controller the public adalAuthenticationService.login() method.

AmitOlkar:
Any updates on fixing this ? My project is also being held back due to this issue.

gfsarri:
Hi,

I implemented the workaround that I posted in Github.

In ui-router, the requireADLogin for the state is set to false.

In the controller (associated to the ui-router state in which I embed powerbi)  I check if there is already a logged in user. If not, then I manually force the login.

adalAuthenticationService.getUser().then(function (user) {      // es una promesa
    console.log(user);
}, function (error) {
    console.log(error);
    adalAuthenticationService.login();
});

Also I had to modify manually adal-angular.js: the returned login method, as shown here:


return {

                    // public methods will be here that are accessible from Controller

                    config: _adal.config,

                    login: function () {

                        _adal.login($location.$$absUrl);      // JAS#

                        // _adal.login();

                    },

                    loginInProgress: function () {

                        return _adal.loginInProgress();


Now it’s working for me. It’s an ugly procedure and I’m looking forward to somebody correcting the root cause of this issue between ui-router and adal-angular. Please, keep me updated if you find out a better solution.

[...]

PR #631, opened 2017-09-28
A hacky fix for #574

mottykohn:
It works but obviously could be better. Just sharing for the interim.

### DIFF
--- a/lib/adal-angular.js
+++ b/lib/adal-angular.js
@@ -65,7 +65,7 @@
 
             // special function that exposes methods in Angular controller
             // $rootScope, $window, $q, $location, $timeout are injected by Angular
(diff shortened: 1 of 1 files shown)


mottykohn:
Did not test if this breaks pre ui-router 1.0.

rohitnarula7176:
@mottykohn Thank you for your contribution. $transitions is not available as a dependency for ui-router versions <1.0. We will need a solution that works for both pre and post ui-router release 1.0.

mottykohn:
@rohitnarula7176 Sorry for late response. I believe I fixed it now.

wy193777:
What's the status for this now? I've spent 2 days on this issue and finally find the problem. Can someone merge it?

wy193777:
Any update?

navyasric:
@wy193777 @rohitnarula7176 is looking into the PR

Issue #636, opened 2017-10-04
Prompt parametr is doubled on renew requests if it's already provided


tamerlan:
Hi,

If you add 'prompt=login' to the extraQueryParameters than the renewal url will contains both 'prompt=login' and 'prompt=none' params and will fail.

rohitnarula7176:
@tamerlan By design , The prompt parameter is not configurable in the library. Can you explain why do you need to add a prompt parameter in the request.

tamerlan:
Hi @rohitnarula7176, to force user to re-enter the login/pass after the logout. Also that fixes the issue #630. You can just remove the users 'prompt' params from the url when generate it for renew.

sorbonad:
The workaround seems to be working on W10/Chrome and W10/Edge. I am still need to do some tests since now I have to enter credentials twice.

rohitnarula7176:
@sorbonad @tamerlan I tried to follow the steps you described above by using 2 users, a corp account and a test account but I am not able to reproduce the issue. I always get token for the user account which I used to sign it. If you add 'prompt=login' to extraQueryParameter property on the config object, adal adds it to all the authorization requests. When you try to send prompt=login in the hidden iframe, adal adds prompt=none on top of it and you will get an error from AAD AADSTS90004: The request is not properly formatted. The parameter &#39;prompt&#39; is duplicated. Can you please specify the version of adal you are using and provide us with a code repro for this issue.

tamerlan:
@rohitnarula7176 You just get the exception which I would like to fix. "The parameter 'prompt' is duplicated". I've created the pull request #645 to fix this.

rohitnarula7176:
@tamerlan Thank you for your contribution. We cannot accept this pull request because your code allows the developers to overwrite the value for prompt in acquireToken methods which we cannot allow by design as it happens in a hidden iframe where prompt cannot be anything other than none. I have made the code change in dev branch such that if you pass prompt value through extraQueryParameter, it will only get added to the login method of adal which is the only one that does not specify prompt value by default. The recommended value for prompt is 'select_account' for login.

PR #645, opened 2017-10-25
Update prompt param if it's already provided in the url (fix fo #636)

tamerlan:
Extracted the function to add parameter to the url. It should prevent duplicated params to appear.

### DIFF
--- a/lib/adal.js
+++ b/lib/adal.js
@@ -434,8 +434,8 @@ var AuthenticationContext = (function () {
         // renew happens in iframe, so it keeps javascript context
         window.renewStates.push(expectedState);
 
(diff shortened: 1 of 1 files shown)


rohitnarula7176:
@tamerlan @tamerlan Thank you for your contribution. We cannot accept this pull request because your code allows the developers to overwrite the value for prompt in acquireToken methods which we cannot allow by design as it happens in a hidden iframe where prompt cannot be anything other than none. I have made the code change in dev branch such that if you pass prompt value through extraQueryParameter, it will only get added to the login method of adal which is the only one that does not specify prompt value by default. The recommended value for prompt is 'select_account' for login.

PR #648, opened 2017-11-01
added silent login feature

rohitnarula7176:


sorbonad:
hello, will this feature allow us to renew the idToken without reloading the page?

PR #650, opened 2017-11-01
Cache bust with changed tenant id

matdarMsft:
When the tenant id changes, ensure that the cached resource for another tenant id is not returned


rohitnarula7176:
@matdarMsft As discussed, this cannot be added as part of the library.

PR #676, opened 2017-12-07
Replaced usage of lambda functions for IE11 support.

zanawar:
Lambda functions aren't currently supported by IE. I've replaced the usage of lambda's with equivalent functions created in the scope of usage to make this library IE compatible again.

For more about lambda function compatibility see [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Browser_compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Browser_compatibility)

### DIFF
--- a/lib/adal-angular.js
+++ b/lib/adal-angular.js
@@ -329,11 +329,11 @@
                     var $transitions = $injector.get('$transitions');
 
                     function onStartStateChangeHandler(transition) {
-                        stateChangeHandler(null, transition.to(), transition.params('to'), transition.from(), transition.params('from'))
+                        stateChangeHandler(null, transition.to(), transition.params('to'), transition.from(), transition.params('from'));
                     }
 
                     function onErrorStateChangeHandler(transition) {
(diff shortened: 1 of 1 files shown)


lib/adal-angular.js
```diff
@@ -327,8 +327,17 @@
 
                 if ($injector.has('$transitions')) {
                     var $transitions = $injector.get('$transitions');
-                    $transitions.onStart({}, (transition) => stateChangeHandler(null, transition.to(), transition.params('to'), transition.from(), transition.params('from')));
-                    $transitions.onError({}, (transition) => stateChangeErrorHandler(null, transition.to(), transition.params('to'), transition.from(), transition.params('from'), transition.error()));
+
+                    function onStartStateChangeHandler(transition) {
+                        stateChangeHandler(null, transition.to(), transition.params('to'), transition.from(), transition.params('from'))
```
rayluo: Nitpicking: would you mind adding a semicolon at the end of this line? Same applies to line 336 too.

rohitnarula7176:
@zanawar Thank you for your contribution. The changes look good. I have merged your pull request.

PR #688, opened 2017-12-18
fix: forwarding request without token to endpoint

DrazenRocket:
There are situations when an application wants to work with endpoint which offer some actions
which require authentication/authorization and some actions which don't.
It's good to forward requests to this type of endpoint because
actions will decide to accept request or not. Without this change (in this pull request), user
must be logged in if he/she wants to send request to an action which doesn't require authentication/authorization of endpoint which is added in initialization of adalAuthenticationServiceProvider.

### DIFF
--- a/lib/adal-angular.js
+++ b/lib/adal-angular.js
@@ -517,8 +517,7 @@
                                     config.headers.Authorization = 'Bearer ' + token;
                                     delayedRequest.resolve(config);
                                 }, function (error) {
-                                    config.data = error;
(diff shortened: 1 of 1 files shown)


stevegomez17:
Forward to @stevegomez17

DrazenRocket:
@stevegomez17 What do you mean when you say "Forward to ..."

DrazenRocket:
@rohitnarula7176 What do you mean? Is this change (pull request) ok?

PR #696, opened 2018-01-03
Add support for scope parameters during authentication.

nhance:
This allows the library to support authorization in addition to authentication, without having to use separate libraries.


lib/adal.js
```diff
@@ -799,7 +796,13 @@ var AuthenticationContext = (function () {
     AuthenticationContext.prototype.promptUser = function (urlNavigate) {
         if (urlNavigate) {
             this.infoPii('Navigate to:' + urlNavigate);
-            window.location.replace(urlNavigate);
+            if (this.config.displayCall) {
```
nhance, line 800: `displayCall` is moved here so we can use it for all navigation. This is useful when we are running in an environment where we don't have access to `window`

PR #704, opened 2018-01-18
Let's use 'renew' in comments to avoid confusion with 'refresh' tokens

snobu:
..since no such thing as a proper refresh token is in use.


rohitnarula7176:
@snobu Thank you for your contribution.

PR #777, opened 2018-07-04
chore: updating the documentation

4F2E4A2E:
@nehaagrawal @rohitnarula7176 @asnow003 @navyasric
No js &/ functionality has been changed in this PR. The build is failing because of a incompatibility with the early node versions (0.10 and 0.11) in the dependencies, though this PR is not intended to address that.


4F2E4A2E:
@nehaagrawal @rohitnarula7176 @asnow003 @navyasric
No js &/ functionality has been changed in this PR. The build is failing because of a incompatibility with the early node versions (0.10 and 0.11) in the dependencies, though this PR is not intended to address that.

4F2E4A2E:
Can we get any updates on this one?

4F2E4A2E:
Why has this one been closed without even a short notice?

PR #786, opened 2018-07-18
Cache the values of _supportsLocalStorage and _supportsSessionStorage

macqm:
Cache the values of `AuthenticationContext.prototype._supportsLocalStorage` and `AuthenticationContext.prototype._supportsSessionStorage` to avoid performance pressure on the browser's local storage APIs.

### DIFF
--- a/lib/adal.js
+++ b/lib/adal.js
@@ -135,6 +135,10 @@ var AuthenticationContext = (function () {
         this._openedWindows = [];
         this._requestType = this.REQUEST_TYPE.LOGIN;
         window._adalInstance = this;
+        this._storageSupport = {
+            localStorage: null,
+            sessionStorage: null
+        };
 
         // validate before constructor assignments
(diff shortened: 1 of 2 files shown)


lib/adal.js
```diff
@@ -1759,37 +1763,52 @@ var AuthenticationContext = (function () {
...
-            if (!window.localStorage) return false; // Test availability
-            window.localStorage.setItem('storageTest', 'A'); // Try write
-            if (window.localStorage.getItem('storageTest') != 'A') return false; // Test read/write
-            window.localStorage.removeItem('storageTest'); // Try delete
-            if (window.localStorage.getItem('storageTest')) return false; // Test delete
-            return true; // Success
+            if (!(storageType in window) || window[storageType] === null) {
+                throw new Error();
+            }
+            var testKey = '__storageTest__';
```
manoj-rath, line 1783: const would have been better here

Issue #682, opened 2017-12-14
Bug: ADAL iFrame's Border is Visible


RehanSaeed:
Immediately after logging in, the iFrame is visible at the bottom of the page because it has a border width of `2px`, you should add a style to remove that. I'm using Bootstrap 4, so maybe the CSS reset doesn't cover this style.

![iframe](https://user-images.githubusercontent.com/2270364/34002428-788b842c-e0ea-11e7-947a-80bd54498642.png)

rohitnarula7176:
@RehanSaeed The iframe border in the code is set to 0px as shown here https://github.com/AzureAD/azure-activedirectory-library-for-js/blob/dev/lib/adal.js#L1677. Can you please explain how it is changed to '2px' in your code.

RehanSaeed:
Not sure what's causing this issue.

Issue #562, opened 2017-05-26
second right scroll bar after renew token


ahadjiGladiator:
I understand that ADAL renew token every hour using hidden iframe.
However whenever this happens it caused second vertical scroll bar.
I am using bootstrap with fixed top and left menu. if I refresh the page (via browser refresh) the second one will disappear. why this is happening?

rohitnarula7176:
@ahadjiGladiator Can you please explain what do you mean by second scroll bar. I will need to know your application set up to investigate this further.

ahadjiGladiator:
I am using angular 1.6 with bootstrap. my page layout has a fix top and left navigation. Below is the code.
My views will be loaded in "<div ng-view></div>".
obviously if page is longer than browser window, I will have right scroll bar. (vertical)
when I use the scroll bar top nav and left nave will stay and content will scroll. (ng-view)
now after 1 hour when access token needs to renew (behind the scene) when I go to another page (like click on a link) a second (vertical) scroll bar will appear in right side.
it is very similar to if you have an ifram inside html and the content is long.
I think this has something to to with adal hidden ifram when renews the access token.   

`<body data-ng-controller="indexCtrl">
    <div class="navbar navbar-default navbar-fixed-top">
        <div class="navbar-header">
            <button type="button" class="navbar-toggle" data-toggle="collapse" data-target=".navbar-collapse">
                <span class="icon-bar"></span>
                <span class="icon-bar"></span>
                <span class="icon-bar"></span>
            </button>
            <a class="navbar-brand active" href="#/"> Admin</a>
        </div>
        <div class="collapse navbar-collapse">
            <ul class="nav navbar-nav">
            </ul>
            <ul class="nav navbar-nav navbar-right" ng-if="userInfo.isAuthenticated">
                <li class="userinfo"><a href="#/">Welcome {{userInfo.profile.given_name}}<i class="fa fa-user fa-fw"></i></a></li>
                <li><a ng-href="#/" data-ng-click="logout()">Logout<i class="fa fa-sign-out fa-fw pull-right"></i></a></li>
            </ul>
        </div>
    </div>
    <div class="row-offcanvas row-offcanvas-left">
        <div id="sidebar" class="sidebar-offcanvas">
            <div class="col-md-12">
                <h3></h3>
                <ul class="nav nav-pills nav-stacked">
                    <li class="nav-divider"></li>
[...]

PR #789, opened 2018-07-27
Fix setting of the borderWidth on adal frame

### LINKED ISSUE #788: adalRenewFrame iFrame's borderWidth not set correctly
It looks like this issues has been previous reported in #682 & #562, but has never been fixed in the library. The root cause is the code setting the borderWidth to "0px" isn't setting it through the element's style property so it is not respected by browsers.

In [adal.js](https://github.com/AzureAD/azure-activedirectory-library-for-js/blob/e9d23e719eccf2341ee38b11323101fb14cb9610/lib/adal.js#L1687) this line should be updated:
```
ifr.style.width = ifr.style.height = ifr.borderWidth = '0px';
```
to
```
ifr.style.width = ifr.style.height = ifr.style.borderWidth = '0px';
```

dlannoye:
This fixes #788, #682, & #562. The root cause was the the code was not correctly setting the borderWidth since it didn't use the style object. Because of this it failed to override default user agent styles provided by the browser causing the iframe to require space.

I also updated a `'` character that was not using the proper encoding in a comment.

### DIFF
--- a/lib/adal.js
+++ b/lib/adal.js
@@ -924,7 +924,7 @@ var AuthenticationContext = (function () {
      * @ignore
      */
     AuthenticationContext.prototype._addHintParameters = function (urlNavigate) {
-        //If you don’t use prompt=none, then if the session does not exist, there will be a failure.
+        //If you don't use prompt=none, then if the session does not exist, there will be a failure.
         //If sid is sent alongside domain or login hints, there will be a failure since request is ambiguous.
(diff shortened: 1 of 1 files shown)


dlannoye:
Any action of this? The CI build failure looks like an issue on your end with the configuration?

rohitnarula7176:
@dlannoye I apologize for the delayed response. Thank you for creating a PR with the fix. I was not able to merge your PR as the build was failing. Since it was a one line change, I have created a PR using the fix you provided. We will include this as part of our next release.

PR #791, opened 2018-08-03
Reworking callback context

dggsax:
The work in this pull request originates from internship project at Microsoft in which we used adal.js for a VSTS Extension. There were issues where context assumptions (primarily that `window.parent === window`) would prevent adal from properly handling window callbacks. Once those assumptions were rearranged, things began to work. Hoping that maintainers can see the changes I made and provide feedback on the substance of them, and if they are aware of other portions within handleWindowCallback() that they can be addressed.

### DIFF
--- a/lib/adal.js
+++ b/lib/adal.js
@@ -924,7 +924,7 @@ var AuthenticationContext = (function () {
      * @ignore
      */
     AuthenticationContext.prototype._addHintParameters = function (urlNavigate) {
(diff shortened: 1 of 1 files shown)


dggsax:
Just noticed the CI Build Failed. Will investigate.

dggsax:
Closing because we are no longer using Adal with our project. Although, the issues proposed should still be investigated with future maintenance of Adal.

Issue #794, opened 2018-08-24
if resource is empty, acquireTokenRedirect calls callback before assigning it


alex3683:
In `adal.js` line [749ff](https://github.com/AzureAD/azure-activedirectory-library-for-js/blob/dev/lib/adal.js#L749-L756):
```js
 AuthenticationContext.prototype.acquireTokenRedirect = function (resource, extraQueryParameters, claims) {
        if (this._isEmpty(resource)) {
            this.warn('resource is required');
            callback('resource is required', null, 'resource is required');
            return;
        }

        var callback = this.callback;
        // ...
```
Here `callback` is called without being set, in case the resource is empty.

rohitnarula7176:
@alex3683 Thank you for pointing out this issue. I apologize for the delayed response. We will release a new version in a couple of weeks with this fix.

rohitnarula7176:
@alex3683 Closing this issue. Fixed and merged in dev.

PR #812, opened 2018-10-03
fixed callback issue with acquireTokenRedirect api

rohitnarula7176:
*Check if callback function exists before calling it.
#794


manoj-rath (approved):
looks good

Issue #806, opened 2018-09-14
JSON.stringify in stateChangeHandler throws on circular reference


ReeseJones:
in [adal-angular.js](https://github.com/AzureAD/azure-activedirectory-library-for-js/blob/dev/lib/adal-angular.js) on line 318 there is the following line:

```
_adal.verbose("State change error occured. Error: " + typeof (error) === 'string' ? error : JSON.stringify(error));
```

The stateChangeErrorHandler  sometimes receives error objects containing circular references which causes the stringify to throw, and obfuscate the errors that are actually happening.

It ends up looking kind of like this.
![image](https://user-images.githubusercontent.com/6557783/45568269-086b3a80-b811-11e8-8fbd-10e025021baa.png)

In our case we had a transition from one state to another while there was an ongoing transition already which produces an error correctly, however it wasn't obvious because because this state change error handler couldnt stringify the error was and throwing in of itself.

rohitnarula7176:
@ReeseJones Thank you for pointing out this issue. I have submitted a PR for this issue. Can you please confirm if it resolves your issue. We will add this fix to our next release in a couple of weeks.

nehaagrawal:
@ReeseJones  Can you please let us know what is the error string that is being passed to JSON.stringify(error) that ended up causing circular reference error?

nehaagrawal:
We have merged the PR. I will update this issue with the version details when we release. Closing this issue for now.

PR #816, opened 2018-10-04
fixed circular reference error in JSON.stringify

rohitnarula7176:
*Fixed bug with JSON.stringify(object) when object has circular reference.
Issue https://github.com/AzureAD/azure-activedirectory-library-for-js/issues/806

### DIFF
--- a/lib/adal-angular.js
+++ b/lib/adal-angular.js
@@ -315,7 +315,7 @@
                 };
 
                 var stateChangeErrorHandler = function (event, toState, toParams, fromState, fromParams, error) {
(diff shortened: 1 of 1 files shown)


lib/adal-angular.js
```diff
@@ -325,6 +325,19 @@
                     }
                 };
 
+                var jsonCircularReferenceReplacer = function () {
```
nehaagrawal, line 328: @rohitnarula7176 can you please add a test case to confirm this is fixed.

nehaagrawal (approved):
I tested it manually by taking few examples of circular reference and it works. It would be good to add a test case.

pkanher617 (approved):

manoj-rath (approved):

PR #818, opened 2018-10-11
Used const instead of var at all the possible places.

sandyjoshi:
Used const instead of var at all the possible places.


sliekens:
Hey I think you should read this (warning: strong language) 
https://jamie.build/const

PR #820, opened 2018-10-16
fix: strict mode nested function adal-angular

Izzmo:
When using the adal-angular library within PhantomJS and strict mode (and in my case Chisel Web Accessibility tool), it throws an error keeping the AngularJS app from running. Moving in the function declaration to an inline call fixes this.

### Tests
Unfortunately, there is no way to test this from what I can tell. If you think you can, please let me know and I'll update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ You can learn in detail about ADAL JS installation and usage documented in the [
 
 ## Versions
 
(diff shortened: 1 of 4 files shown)


m4m4m4:
UglifyJS also renders an error due to these nested functions

PR #839, opened 2018-11-30
ADAL release 1.0.18

nehaagrawal:
ADAL release 1.0.18


manoj-rath:
we need to remove the jsonCircularReferenceReplacer PR because of the above comment by abub001, we can fix that and merge separately.

m4m4m4:
This is taking forever..

PR #840, opened 2018-12-04
Reverse the #839 Circular Reference Error Fix - currently customer reported a scope error

sameerag:
Reverting the change for further inspection as a user reported "jsonCircularReferenceReplacer() is undefined in the responseError of AdalModule.factory due to the scope it is placed at here."


manoj-rath (approved):
need to make sure dev branch is clean while we work on the correct fix.

Issue #764, opened 2018-06-11
Refresh token problem. Token expiring after 1 hour.


viganella:
Hi 
When reading adal documentation I saw that when user authenticates to AAD they have an access_token and a refresh token (that can renew acccess_token for up to 90 days).
I have an application that uses websocket connection to get data from cloud and after 1 hour letting it run I get 'access_token_expired ' message.
My question is that, is there anything special i have to do to let user stay logged-in for a longer period of time e.g. for days ? 
Or is this another bug?

viganella:
After 1 hour I get the error message ('user login is required').
I have 'keep my signed in option enabled'.
My localstorage:
![token_expired_localhoststorage](https://user-images.githubusercontent.com/2545516/41313034-1d5b2634-6e89-11e8-8305-05ce7349d3e9.png)

viganella:
I have investigated further here and what seems to have caused that is because on refresh I was loosing (this_user <-null), so that after 1 hour Adal was requesting for a fresh log-in.
This is not ok of course, since I should not loose user after he refreshes, it is just very unreliable.

viganella:
Following up I actually am calling
this.get('session')._context.getCachedUser();  ON page REFRESH 
which will actually set again a new user.

mchint01:
We are having same error, what is the resolution for this?
We are calling getCachedUser() on page refresh as well. Still having the issue of token not renewing

atlefren:
Can also reproduce this bug. Can confirm it is related to this._user beeing null after a page reload. Will submit a PR with fix asap

sameerag:
Please check documentation on token renewal [here](https://github.com/AzureAD/microsoft-authentication-library-for-js/wiki/FAQs#q3-how-to-renew-tokens-with-msaljs). All tokens expire after an hour, but calling `acquireTokenSilent()` in the background before the token expiry renews the token.

All current authentication work from microsoft is delivered through `msal js` library [here](https://github.com/AzureAD/microsoft-authentication-library-for-js). `adal js` is still supported only for security fixes. We would recommend to move to `msal js` for any advanced feature asks.

atlefren:
So, after 1.5 years this issue is closed with a link to how to do the asked thing in msal. Thus: for all intents and purposes adal.js is deprecated? 

Could you please state this clearly in the readme?

sameerag:
Yes, we will soon. We are trying to figure out usage issues vs feature asks and will add the documentation soon.

PR #857, opened 2019-03-27
Call getCachedUser in acquireToken to avoid bugs related to renew 

atlefren:
Directly calling this._user in acquireToken will result in a "User login is required" error if the page has been reloaded. 

The fix is to call getCachedUser, so that user is loaded from cache.

This solves #764

PR against master as this is a pressing issue that should be fixed asap. Please advice if I should use another branch.

### DIFF
--- a/lib/adal.js
+++ b/lib/adal.js
@@ -926,7 +926,7 @@ var AuthenticationContext = (function () {
      * @ignore
      */
     AuthenticationContext.prototype._addHintParameters = function (urlNavigate) {
-        //If you don�t use prompt=none, then if the session does not exist, there will be a failure.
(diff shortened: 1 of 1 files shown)


atlefren:
Hmm, no idea why tests fail: they work fine on my machine:

```
atlefren@atlsve-M4800:~/code/azure-activedirectory-library-for-js [issue_764]$ npm test

> adal-angular@1.0.17 test /mnt/d/code/azure-activedirectory-library-for-js
> jasmine-node tests/unit/spec

.........Waiting for initial timeout
.Waiting for initial timeout
.............................Waiting for initial timeout
..................

Finished in 2.355 seconds
57 tests, 168 assertions, 0 failures, 0 skipped
```

PR #898, opened 2020-02-27
Fix issue that ADAL mistakenly set token expiration time in local storage

Gang-Peng:
What:
ADAL has this issue that it will update 'adal.expiration.key' based on current local time + expires_in info from AAD, however, if the callback is not executed on time due to all kinds of reasons, such as browser tabs are inactive or computer is sleeping, then it may get into this state that token expiration time is mistakenly set.

How:
To fix the issue, when ADAL requests AAD access token, it will pass its current local time in the "state" query parameter, which will be returned by AAD on success, and we will use the passed in local time in "state" to calculate the correct token expiration time and set it to "adal.expiration.key" inside local storage.


lib/adal.js
```diff
@@ -1153,16 +1153,31 @@ var AuthenticationContext = (function () {
...
+     */
+    AuthenticationContext.prototype._getExpireBaseTimeFromState = function (state) {
+        if (state) {
+            var splittedInfo = state.split('|');
+
+            if (splittedInfo.length > 2) {
+                return parseInt(splittedInfo[2], 10);
+            }
+        }
+        return 0;
```
jasonnutter: Could this instead default to the current time (which would simplify the downstream logic)?

jasonnutter:
Thanks @Gang-Peng, this is a great fix!

lib/adal.js
Gang-Peng: fixed as suggested

PR #911, opened 2020-05-25
Fix border-width style on hidden adal frame

polys:
When `borderWidth` isn't set properly, the browser applies a default
non-zero width which then interferes with the page layout. The frame
wasn't truly 'hidden'.


jasonnutter:
Closing, as we will be arching this repo and moving the ADAL.js code to the MSAL.js repo: https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/maintenance/adal-angular

Please reopen the PR there, thanks!

PR #915, opened 2020-06-26
[Yammer] improve logout experience in multiple tabs of the same application

jessemartin:
Problem overview
---
Currently, if a user has open multiple browsing contexts (tabs, windows, etc) of an application that uses ADAL, logout does not sync across the tabs. When `logOut` is called in one tab, others continue to maintain `this._user` in memory. Therefore, calls to `getUser` and `getCachedUser` in these tabs will continue to return a user object, even though the localStorage/sessionStorage entries have been removed and the user should be considered "logged out".

Effect on Yammer
---
Yammer is using an authenticator that checks for user presence as part of determining if a login is required during token acquisition. When a logout has occurred in one tab, and acquireToken is then called in another tab, the second tab would still be considered "logged-in", because `getCachedUser` still returns a user. As a result, acquireToken would be called when no `login.microsoftonline.com` cookies exist, causing ADAL to return `AADSTS50058` (cookies not found error) and Yammer to display a third party cookie configuration error page.

Proposed fix
---
This PR adds a mechanism to clear the user when an external browsing context of the same origin performs a logout. The approach uses [storage events](https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event) instead of something like the [broadcast channel api](https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API) to maintain compatibility with browsers supported by Yammer (specifically IE11 and Safari).

### DIFF
--- a/lib/adal.js
+++ b/lib/adal.js
@@ -187,6 +187,8 @@ var AuthenticationContext = (function () {
         if (this.config.loadFrameTimeout) {
             this.CONSTANTS.LOADFRAME_TIMEOUT = this.config.loadFrameTimeout;
         }
+
+        this._setupCrossBrowsingContextLogoutCleanup();
     };
 
     if (typeof window !== 'undefined') {
@@ -855,7 +857,7 @@ var AuthenticationContext = (function () {
      */
     AuthenticationContext.prototype.logOut = function () {
         this.clearCache();
-        this._user = null;
+        this._clearUser();
         var urlNavigate;
 
         if (this.config.logOutUri) {
@@ -982,6 +984,14 @@ var AuthenticationContext = (function () {
         }
 
         return user;
+    };
+
+    /**
+     * Clears the user
+     * @ignore
+     */
+    AuthenticationContext.prototype._clearUser = function () {
+        this._user = null;
     };
 
     /**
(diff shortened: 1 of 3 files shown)


tests/unit/spec/AdalSpec.js
```diff
@@ -1020,11 +1039,9 @@ describe('Adal', function () {
     });
 
     it('tests default value of redirect uri', function () {
-        global.window = {
```
jessemartin, line 1023: This was wiping out fake functions set on window.

lib/adal.js
```diff
@@ -1793,6 +1803,22 @@ var AuthenticationContext = (function () {
...
+    AuthenticationContext.prototype._setupCrossBrowsingContextLogoutCleanup = function () {
+        var clearUserOnExternalBrowsingContextLogout = function (evt) {
+            var isExternalBrowsingContextLogout = evt.key === this.CONSTANTS.STORAGE.IDTOKEN && evt.oldValue !== '' && evt.newValue === '';
+
+            if (isExternalBrowsingContextLogout) {
+                this._clearUser();
+            }
+        };
+
+        window.addEventListener('storage', clearUserOnExternalBrowsingContextLogout.bind(this));
```
Gang-Peng, line 1819: are we listening to all the storage change events now? Will that be a lot?

Gang-Peng: should we create a different local storage key just for this logout scenario? Instead of relying on this.CONSTANTS.STORAGE.IDTOKEN ? Will that be more clear?

jessemartin: Yep, AFAIK, that's how `storage` listeners must work. Then the handler function is responsible for acting only on the events that are deemed appropriate. 

I considered creating a new storage key for this, but the clearing of id token was pre-existing and does exactly what we want, so I figure it would be okay to use that instead of adding another storage entry and having to clean that up.

jasonnutter, line 1819: This works on IE and Safari? MDN doesn't have any info: https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event#Browser_compatibility

jasonnutter: I suppose not a big deal if not.

jessemartin: I tested it out in Safari 13 and IE11, both of which seem to work.

PR #917, opened 2020-07-06
use self._activeRenewals to decide whether we should trigger timeout

Gang-Peng:
use self._activeRenewals to decide whether we should trigger timeout clean ups, instead of 'self.CONSTANTS.STORAGE.RENEW_STATUS + resource' since it could be changed by other tabs.

### DIFF
--- a/lib/adal.js
+++ b/lib/adal.js
@@ -589,14 +589,14 @@ var AuthenticationContext = (function () {
         var self = this;
 
         setTimeout(function () {
-            if (self._getItem(self.CONSTANTS.STORAGE.RENEW_STATUS + resource) === self.CONSTANTS.TOKEN_RENEW_STATUS_IN_PROGRESS) {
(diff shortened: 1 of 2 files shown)


lib/adal.js
```diff
@@ -589,14 +589,14 @@ var AuthenticationContext = (function () {
         var self = this;
 
         setTimeout(function () {
-            if (self._getItem(self.CONSTANTS.STORAGE.RENEW_STATUS + resource) === self.CONSTANTS.TOKEN_RENEW_STATUS_IN_PROGRESS) {
-                // fail the iframe session if it's in pending state
+            if (self._activeRenewals[resource]) {
+                // fail the iframe session when times out and there are active toke renewals
```
jessemartin: `token` :)

Gang-Peng: fixed

jasonnutter (approved):

PR #926, opened 2021-01-19
fixing timing issue with redirect

moonchoon:
this will cancel out any previous schedule attempt to redirect and let handleWindowCallback to redirect to correct originating url.

Issue - 
when we trying to attempt silent signin with "prompt=none" (with redirect), first attempt work fine and login.microsoft.com first redirect to teams.microsoft.com with error_description in url hash and then adal.js redirect browser to originating url i.e. teams.microsoft.com (expected behavior)

Now, if we refresh page, again teams.microsoft.com try to silent signin with prompt=none, but when login.microsoftonline.com redirect back to teams.microsoft.com with error_description in url hash, adal.js instead of redirecting back to originating url, it redirect again to login.microsoftonline.com without prompt=none.

This seems like some scheduling of redirect to login.microsoftonline.com. So even with location.href=<originating url>, it always redirect to login.microsoftonline.com.

example:
setTimeout(function(){window.location.href="https://www.msn.com"},100);window.location.href="https://www.google.com";

proposed fix, will make sure to redirect to expected url and resolve any timing issue with href change.

### DIFF
--- a/lib/adal.js
+++ b/lib/adal.js
@@ -1362,7 +1362,9 @@ var AuthenticationContext = (function () {
 
             if (window.parent === window && !isPopup) {
                 if (self.config.navigateToLoginRequestUrl) {
-                    window.location.href = self._getItem(self.CONSTANTS.STORAGE.LOGIN_REQUEST);
+                    var navigateUrl = self._getItem(self.CONSTANTS.STORAGE.LOGIN_REQUEST);
+                    window.setTimeout(function(){window.location.href = navigateUrl;}, 0);
+                    window.location.href = navigateUrl;
                 } else window.location.hash = '';
             }
         }


jasonnutter:
> Now, if we refresh page, again teams.microsoft.com try to silent signin with prompt=none, but when login.microsoftonline.com redirect back to teams.microsoft.com with error_description in url hash, adal.js instead of redirecting back to originating url, it redirect again to login.microsoftonline.com without prompt=none.

I'm not sure the proposed solution is the appropriate way to fix this issue. `CONSTANTS.STORAGE.LOGIN_REQUEST` is always set to a location on the application (e.g. `teams.microsoft.com`), not to a url on `login.microsoftonline.com`. If navigating to the location saved as `CONSTANTS.STORAGE.LOGIN_REQUEST` results in the user ending up on `login.microsoftonline.com` (with no `prompt=none` set in the url), that likely means something else is causing that undesired redirect, e.g. your application invoking loginRedirect/acquireTokenRedirect.

lib/adal.js
```diff
@@ -1362,7 +1362,9 @@ var AuthenticationContext = (function () {
 
             if (window.parent === window && !isPopup) {
                 if (self.config.navigateToLoginRequestUrl) {
-                    window.location.href = self._getItem(self.CONSTANTS.STORAGE.LOGIN_REQUEST);
+                    var navigateUrl = self._getItem(self.CONSTANTS.STORAGE.LOGIN_REQUEST);
+                    window.setTimeout(function(){window.location.href = navigateUrl;}, 0);
```
jasonnutter, line 1366: I'm not sure I understand why this approach was choosen. Is intent that this will delay the redirecting to `CONSTANTS.STORAGE.LOGIN_REQUEST`? With the underlying issue being that the redirect is happening too early, thus causing the app to send the user back to `login.microsoftonline.com`?

jasonnutter:
@pkanher617 @sameerag Any thoughts?

lib/adal.js
moonchoon: This will clear any request pending for login.microsoftonline.com and user will be redirect correctly to CONSTANTS.STORAGE.LOGIN_REQUEST

jasonnutter: `self.CONSTANTS.STORAGE.LOGIN_REQUEST` should never be set to a `login.microsoftonline.com` url. This url is always set to `window.location.href` when the user is on your application (e.g. `teams.microsoft.com`), right before the redirect to `login.microsoftonline.com`:

- login: https://github.com/AzureAD/azure-activedirectory-library-for-js/blob/3c9ba88a63d07ea081364141946216f7103261f4/lib/adal.js#L225
- acquireToken: https://github.com/AzureAD/azure-activedirectory-library-for-js/blob/3c9ba88a63d07ea081364141946216f7103261f4/lib/adal.js#L791

moonchoon: Agree. but somewhere redirect to login.microsoftonline.com is pending which redirect browser to login.microsoftonline.com instead of CONSTANTS.STORAGE.LOGIN_REQUEST.

This approach which make sure to clear any request pending for login.microsoftonline.com and user will be redirect correctly to CONSTANTS.STORAGE.LOGIN_REQUEST

jasonnutter:
Closing, as we will be arching this repo and moving the ADAL.js code to the MSAL.js repo: https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/maintenance/adal-angular

Please reopen the PR there, thanks!

PR #933, opened 2021-06-02
correctly set iframe border width to 0

jessemartin:
The `borderWidth` attribute was being set directly on the iframe element, when it should instead be set on the iframe's `style` object.

This bug enabled the token refresh iframe to take up space when borderWidth was not globally set to 0, resulting in elements shifting and extra scrollbars being created.


jessemartin:
@jasonnutter mind reviewing?

